PR #1, opened 2025-09-17
ci: fix failing tests.

vytautas-astrauskas-sensmetry:


tilowiklundSensmetry (approved):
:+1: Works for now

We should consider mimicking `uv` in the future (with `--offline` e.t.c.).

I'd also like to disable index throughout the integration tests, but let's handle that at the config file level rather than CLI.

Issue #13, opened 2025-09-19
Allow indicating certain projects (IRIs) to be marked as provided elsewhere


tilowiklundSensmetry:
We should have a mechanism for indicating that certain IRIs are provided elsewhere, and should thus not be listed/installed .

The main (current) use is to allow excluding standard library projects from being installed, as these would often be included in the tooling (e.g. in syside).

PR #15, opened 2025-09-19
Simple temporary work-around for skipping standard library installation

tilowiklundSensmetry:
Default to not installing any of the known IRIs of the KerML/SysML Beta1 standard libraries. Adds a flag `--include-std` to override this behaviour. The standard library is still treated normally when solving for dependencies, but is excluded during actual installation.

I've only done manual testing for this functionality, as it is only a short term temporary work-around until https://github.com/sensmetry/sysand/issues/13 is treated properly.

### DIFF
--- a/sysand/src/cli.rs
+++ b/sysand/src/cli.rs
@@ -40,7 +40,13 @@ pub enum Command {
         command: Option<EnvCommand>,
     },
     /// Sync env to lockfile
-    Sync,
+    Sync {
+        /// Includes the KerML/SysML standard libraries when installing.
+        /// By default these are excluded because they are typically
+        /// shipped with your language implementation.
+        #[arg(long, default_value = "false")]
+        include_std: bool,
+    },
     /// Prints the root directory of the current project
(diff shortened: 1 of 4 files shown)


sysand/src/cli.rs
```diff
@@ -40,7 +40,11 @@ pub enum Command {
         command: Option<EnvCommand>,
     },
     /// Sync env to lockfile
-    Sync,
+    Sync {
+        /// Excludes the KerML/SysML standard libraries from being installed
+        #[arg(long, default_value = "true")]
+        include_std: bool,
```
vytautas-astrauskas-sensmetry, line 46:
```suggestion
        /// Includes the KerML/SysML standard libraries when installing.
        /// By default these are excluded because they are typically
        /// shipped with your language implementation.
        #[arg(long, default_value = "false")]
        include_std: bool,
```

vytautas-astrauskas-sensmetry:
For me still installs the standard library. However, I do not see why.

PR #20, opened 2025-09-23
Add support for multiple indices

tilowiklundSensmetry:
Introduces `SequentialResolver` for running multiple resolvers (of the same type) in sequence, and uses this from the CLI and Python bindings to allow multiple indices.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3297,7 +3297,7 @@ dependencies = [
 
 [[package]]
 name = "sysand"
-version = "0.0.1"
+version = "0.0.2"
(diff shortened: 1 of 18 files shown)


victor-linroth-sensmetry:
> It's a difference in intent, at the moment there is no difference in behaviour, but we may well want to (at the very least) take it into account when generating error messages.

Without a concrete example it just sounds like rubbish to me, but time will I guess.

victor-linroth-sensmetry (approved):
Seems fine for now. I'll try to maybe add some config options at a later point.

PR #22, opened 2025-09-23
Make `lock` and `sources` properly take into account excluded IRIs

tilowiklundSensmetry:
sysand (internally) supports marking certain projects as being provided, meaning they're expected to be made available by your language implementation or some other external mechanism. Currently this functionality is only exposed with the `--include-std` flag, without which a hard-coded (current) set of KerML/SysML standard libraries are assumed to be provided.

This MR makes sysand take into account the provided projects during `lock`. When solving dependencies (usages), IRIs that have *some* provided version may only be satisfied by one of the provided versions. In other words, provided projects completely override the corresponding IRI.

It also makes sysand take into account the provided projects during `sources`. When listing source files provided projects are simply ignored, but assumed to be present.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[target.'cfg(windows)']
+rustflags = ["-Ctarget-feature=+crt-static"]


core/src/commands/sync.rs
```diff
@@ -30,6 +32,14 @@ pub enum SyncError<UrlError> {
     UnsupportedSources(String),
     #[error("failed to install project {uri}:\n{cause}")]
     InstallFailure { uri: String, cause: String },
+    #[error(
+        "tried to install {iri} (checksum {hash}) which is not among the versions supported by your environment"
```
vytautas-astrauskas-sensmetry, line 36: What is “environment” in this context? Is it the things in `sysand_env` or the tool like Syside?

core/src/resolve/replace.rs
```diff
@@ -0,0 +1,290 @@
+use std::io::Read;
+
+use thiserror::Error;
+
+use crate::{project::ProjectRead, resolve::ResolveRead};
+
+/// Resolver that overrides the resolution of some underlying (secondary)
+/// by that of another (primary) resolver.
```
vytautas-astrauskas-sensmetry, line 8: Names `primary` and `secondary` tell nothing about the intended use of these fields. Maybe we could rename them to, for example, `replaced` and `replacing`?

vytautas-astrauskas-sensmetry: Actually, after reading other code, I think that `Replace` in `ReplaceResolver` is the confusing part here. As far as I understand, it is not replacing, but instead `ResolverWithFallback` that tries `primary` and if it fails, tries `secondary`?

vytautas-astrauskas-sensmetry: Or maybe `PriorityResolverSequence` as that is how you named the test?

bindings/py/src/lib.rs
```diff
@@ -201,14 +202,21 @@ fn do_build_py(output_path: String, project_path: Option<String>) -> PyResult<()
...
 )]
 pub fn do_sources_env_py(
     env_path: String,
     iri: String,
     version: Option<String>,
     include_deps: bool,
+    include_std: bool,
 ) -> PyResult<Vec<String>> {
+    let provided_iris = if !include_std {
+        known_std_libs()
```
vytautas-astrauskas-sensmetry, line 215: I think that bindings must have a way to provide the list of libraries as an argument because the tools know best what libraries they have bundled inside them.

No need to address this point in this PR, but then we need to have an issue to not forget this point.

tilowiklundSensmetry: There is already an issue for this (the same behaviour should be available in the CLI), the current behaviour is mentioned as a temporary stop gap in the original MR.

https://github.com/sensmetry/sysand/issues/13

core/src/commands/sync.rs
tilowiklundSensmetry: Agreed, confusing use of words, rephrased it.

core/src/resolve/replace.rs
tilowiklundSensmetry: Renamed `replace[...]` to `priority[...]` and `primary` to `higher` and `secondary` to `lower.

vytautas-astrauskas-sensmetry: Not perfect, but good enough for now.

vytautas-astrauskas-sensmetry (approved):

PR #36, opened 2025-10-06
Make Java bindings inherit metadata values from workspace Cargo.toml

### LINKED ISSUE #35: Java bindings do not inherit Cargo metadata
`bindings/py/Cargo.toml` and `bindings/js/cargo.toml` inherit Cargo workspace metadata, like version, authors, etc.

`bindings/java/Cargo.toml` does not do this, but it should.

andrius-puksta-sensmetry:
Closes #35.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3361,7 +3361,7 @@ dependencies = [
 
 [[package]]
 name = "sysand-java"
-version = "0.0.1"
+version = "0.0.2"
 dependencies = [
  "indexmap",
  "jni",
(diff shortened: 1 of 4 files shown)


andrius-puksta-sensmetry:
Ok, so Java bindings versioning (prior to this PR) works as follows:
- Java bindings Maven package inherits version from Rust crate `sysand-java` via `java-builder.py`
- Java bindings test Maven package has a hardcoded dependency on Java bindings 0.0.1

So for this to work, we have these options:
1. keep Java bindings Rust crate at version 0.0.1 (other bindings inherit versions from cargo workspace)
2. separately version Java bindings Maven package and keep version in sync with test package
3. inherit Rust package version in test package dependency

I would go with option 3 at least for now.

PR #39, opened 2025-10-07
Make Java build Python script executable

andrius-puksta-sensmetry:
This allows running it directly: `./java-builder.py`.


vytautas-astrauskas-sensmetry (approved):
Thanks!

PR #74, opened 2025-10-15
First version of async traits and async HTTP

tilowiklundSensmetry:
This introduces new async traits `ProjectReadAsync`, `ReadEnvironmentAsync`, and `ResolveReadAsync`; mirroring `ProjectRead`, `ReadEnvironment`, and `ResolveRead`.

All `reqwest`-based implementations have been changed to async, using `reqwest_middleware::ClientWithMiddleware` instead of `reqwest::blocking::Client`. 

Utility classes such as `SequentialResolver` should now work for both synchronous and asynchoronous use, i.e. `SequentialResolver<T>` is `ResolveReadAsync` if `T: ResolveReadAsync` and is `ResolveRead` if `T: ResolveRead`.  To the extent possible, they try to intelligently `join` futures to allow concurrency.

Utility classes `AsAsync[Project|Environment|Resolver]<T>` and `AsSync[Project|Environment|Resolver]Tokio<T>` allow treating a synchronous implementation as a (trivially) asynchronous one, and to treat an asynchronous one as if it were synchronous (if provided with a `tokio` runtime). Traits functions `to_async` and `to_sync_tokio` have been added to all the relvant traits. 

The CLI utilises the new async version of HTTP implementations, at the cost of having to pass around an `Arc<tokio::runtime::Runtime>`. This is a bit messy, but, I think, good enough for the moment. 

NOTE: I have (for now) removed support for range requests when accessing kpar archives. In practice this was (at the moment) slower, and I couldn't find a readily available, simple, async way of implementing it.

### DIFF
--- a/core/src/project/mod.rs
+++ b/core/src/project/mod.rs
@@ -149,7 +149,7 @@ pub trait ProjectRead {
         Ok(self.get_info()?.map(|info| info.name))
     }
 
-    /// `is_definitely_invalid` will return `true`` only if get_project() would definitely
+    /// `is_definitely_invalid` will return `true` only if get_project() would definitely
     /// produce an error or return `Some((info, meta))` where either `info` or `meta`
     /// are `None`. If it returns `false` nothing definite can be said.
     ///
--- a/bindings/js/package-lock.json
+++ b/bindings/js/package-lock.json
@@ -12,15 +12,15 @@
         "copy-webpack-plugin": "^13.0.1",
         "glob": "^11.0.3",
         "jasmine-browser-runner": "^3.0.0",
-        "jasmine-core": "^5.12.0",
+        "jasmine-core": "^5.12.1",
         "webpack": "^5.102.1",
         "webpack-cli": "^6.0.1"
       }
     },
     "node_modules/@bazel/runfiles": {
-      "version": "6.3.1",
-      "resolved": "https://registry.npmjs.org/@bazel/runfiles/-/runfiles-6.3.1.tgz",
-      "integrity": "sha512-1uLNT5NZsUVIGS4syuHwTzZ8HycMPyr6POA3FCE4GbMtc4rhoJk8aZKtNIRthJYfL+iioppi+rTfH3olMPr9nA==",
+      "version": "6.5.0",
+      "resolved": "https://registry.npmjs.org/@bazel/runfiles/-/runfiles-6.5.0.tgz",
+      "integrity": "sha512-RzahvqTkfpY2jsDxo8YItPX+/iZ6hbiikw1YhE0bA9EKBR5Og8Pa6FHn9PO9M0zaXRVsr0GFQLKbB/0rzy9SzA==",
       "dev": true,
       "license": "Apache-2.0"
     },
@@ -62,6 +62,7 @@
       "resolved": "https://registry.npmjs.org/@isaacs/cliui/-/cliui-8.0.2.tgz",
       "integrity": "sha512-O8jcjabXaleOG9DQ0+ARXWZBTfnP4WNAqzuiJK7ll44AmxGKv/J2M4TPjxjY3znBCfvBXFzucm1twdyFybFqEA==",
       "dev": true,
+      "license": "ISC",
       "dependencies": {
         "string-width": "^5.1.2",
         "string-width-cjs": "npm:string-width@^4.2.0",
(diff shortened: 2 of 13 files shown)


tilowiklundSensmetry:
> Skimmed through. I would need to start using async more extensively to see whether the design works since I do not have much experience with it. I would suggest merging and if needed changing things later.
> 
> The only thing that I did not get is why there is much commented out code. Is all of it related to range requests?

This is still effectively a draft merge request.

tilowiklundSensmetry:
Ok, this appears to be mostly working now, I will try to fix the bindings and address comments.

core/src/env/mod.rs
```diff
@@ -80,6 +85,206 @@ pub trait ReadEnvironment {
...
+            )
+            .await
+            .into_iter()
+            .collect()
+        }
+    }
+
+    // Maybe make this return an associated type instead? Would, for example, allow
+    // .as_async.as_tokio_sync == .as_tokio_sync.as_async == id
+    fn to_tokio_sync(self, runtime: Arc<tokio::runtime::Runtime>) -> AsSyncEnvironmentTokio<Self>
```
andrius-puksta-sensmetry: Are there any downsides to returning an associated type? If no, the let's do it.

core/src/env/mod.rs
```diff
@@ -80,6 +85,206 @@ pub trait ReadEnvironment {
...
+    {
+        AsSyncEnvironmentTokio {
+            runtime,
+            inner: self,
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct AsAsyncEnvironment<T> {
```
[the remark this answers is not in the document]
tilowiklundSensmetry: done

core/src/project/mod.rs
```diff
@@ -350,6 +558,126 @@ pub trait ProjectMut: ProjectRead {
...
+pub struct AsAsyncProject<T> {
+    pub inner: T,
+}
+
+#[derive(Debug)]
+pub struct AsAsyncReader<T> {
+    inner: T,
+}
+
+impl<T: Read + std::marker::Unpin> AsyncRead for AsAsyncReader<T> {
```
andrius-puksta-sensmetry: Suggestion: import and use at least known std types directly. Less clutter, and fully qualified paths arguably don't add clarity, especially for the likes of `String`, `Sync`, `Pin` and similar. This of course applies to the whole project.

core/src/project/reqwest_kpar_download.rs
```diff
@@ -80,11 +92,17 @@ impl ReqwestKparDownloadedProject {
...
-        .map_err(|e| FsIoError::WriteFile(self.inner.archive_path.to_path_buf(), e))?;
+        while let Some(bytes) = bytes_stream.next().await {
+            let bytes = bytes.map_err(|e| {
+                ReqwestKparDownloadedError::Reqwest(self.url.as_str().to_string().into(), e.into())
+            })?;
+            file.write_all(&bytes)
+                .map_err(|e| FsIoError::WriteFile(self.inner.archive_path.to_path_buf(), e))?;
+        }
 
         file.flush()
```
andrius-puksta-sensmetry: `flush()` and other file I/O can be slow. Maybe [tokio's file](https://docs.rs/tokio/latest/tokio/fs/struct.File.html) should be used here and in other async functions?

.gitignore
```diff
@@ -3,3 +3,4 @@ docs/book/
 sysand-py/python/sysand/_sysand_core.abi3.so
 **/__pycache__
 .DS_Store
+.vscode/
```
andrius-puksta-sensmetry: No newline at end of file :)

andrius-puksta-sensmetry: Recommend configuring the editor to always do it automatically. Maybe this was some tool though.

core/src/env/mod.rs
tilowiklundSensmetry: I already use the sync <-> async conversion in the main implementation.

sync -> async is useful mostly when assembling resolvers (such as when using `StandardResolver`) and some of the constituent resolvers are synchronous, while the overall resolver is asynchronous in order to utilise the asynchronous nature of some of the other constituents.

async -> sync means you can more easily control the boundary at which asynchronous behaviour starts.

core/src/project/reqwest_kpar_download.rs
tilowiklundSensmetry: Let's make it a future MR though, this one is already big enough and we can easily improve the IO later.

PR #80, opened 2025-10-20
Fix moving files/directories between mount points

andrius-puksta-sensmetry:
Logic:
- try `std::fs::rename`
- if it succeeds, done
- if it fails:
  - if src is file, copy file
  - if src is dir, recursively copy it
- delete src


andrius-puksta-sensmetry:
Fixed and tested locally.

tilowiklundSensmetry (approved):
Seems to work!

Issue #33, opened 2025-10-02
Produce explanatory error messages on lexer/parser failure


tilowiklundSensmetry:
Currently lexing/parsing errors produce very short errors, without reference to the source code. This means that `sysand include` and `sysand build` may fail in ways that are hard to understand and debug for the user.

Error messages should at a minimum show the offending part of the code, in which file it appears, and ideally contain a longer, human readable, explanation of the error.

PR #87, opened 2025-10-24
refactor(all)!: add more details to errors

andrius-puksta-sensmetry:
### Core

- Added a general error type `FsIoError`, which consolidates all filesystem I/O errors and includes relevant context. It is included in every error type that needs to represent filesystem errors.
- Added relevant details to every error type in core.
- Create `mod wrapfs` that wraps `std`'s I/O methods that we use and adds relevant context automatically
- Box more large errors
- Allow unused imports in tests (`#![allow(unused_imports)]`) to silence warnings. These imports are used by different features and it's not worth it figuring out where exactly they are used.
- Implements #33. It now prints the name of the file that failed to parse, tries to give some explanation and mostly prints the offending piece of code. Still doesn't give the line numbers.

### Java bindings

- More safety for exceptions: add `enum ExceptionKind` that lists all possible Sysand exceptions and use it instead of providing string names for exceptions. Fixed one case of nonexistent exception name.
- Make Java bindings and their tests actually work on Java 8 and enforce it in their `pom.xml`.
- More ergonomic exception throwing: instead of providing `&mut env` on each call, implement a trait on `env` directly

### Python bindings

- Replaced deprecated methods with their exact counterparts (deprecated methods just called them under the hood).

### WASM bindings

- Clear `localStorage` on start of one test (the test uses less storage slots than the other one, and so entries form the other test are still visible). Maybe should clear `localStorage` at the end of every test?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,15 +3,6 @@
 version = 4
 
 [[package]]
-name = "addr2line"
-version = "0.24.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
-dependencies = [
- "gimli",
-]
-
-[[package]]
 name = "adler2"
 version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -30,9 +21,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.3"
+version = "1.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8e60d3430d3a69478ad0993f19238d2df97c507009a52b3c10addcd7f6bcb916"
+checksum = "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301"
 dependencies = [
  "memchr",
 ]
@@ -54,9 +45,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.20"
+version = "0.6.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3ae563653d1938f79b1ab1b5e668c87c76a9930414574a6583a7b7e11a8e6192"
+checksum = "43d5b281e737544384e969a5ccad3f1cdd24b48086a0fc1b2a5262a26b8f4f4a"
 dependencies = [
  "anstyle",
  "anstyle-parse",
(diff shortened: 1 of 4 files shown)


core/src/project/reqwest_src.rs
```diff
@@ -103,15 +103,25 @@ impl ProjectRead for ReqwestSrcProject {
     > {
         let info_resp = self.get_info().send()?;
         let meta_resp = self.get_meta().send()?;
-
         let info: Option<InterchangeProjectInfoRaw> = if info_resp.status().is_success() {
-            Some(serde_json::from_str(&info_resp.text()?)?)
+            // FIXME: find a way to not clone this in all cases
+            // see https://github.com/seanmonstar/reqwest/issues/1542
+            let url = info_resp.url().to_string();
```
andrius-puksta-sensmetry: This is certainly not ideal, but we have to clone this to be able to use it in a potential error, because `info_resp.text()` consumes `info_resp` and there is no way to get both url and text from `info_resp`. Any better ideas?

core/src/symbols/lex.rs
andrius-puksta-sensmetry: @vytautas-astrauskas-sensmetry why are these not parsed using the attributes?

bindings/js/tests/basic_browser.rs
```diff
@@ -73,9 +73,11 @@ mod browser_tests {
 
     #[wasm_bindgen_test(unsupported = test)]
     fn test_basic_env() -> Result<(), Box<dyn Error>> {
-        do_env_js_local_storage("sysand_storage", "/").unwrap();
-
         let local_storage = web_sys::window().unwrap().local_storage().unwrap().unwrap();
+        // Local storage is not automatically cleared between tests.
+        local_storage.clear().unwrap();
```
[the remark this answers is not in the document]
tilowiklundSensmetry: Probably a good idea.

core/src/project/reqwest_src.rs
tilowiklundSensmetry: I doubt this will ever be a bottleneck, so I don't think it's a problem.

tilowiklundSensmetry:
> Is it worth putting more work into addressing #33? Our sysml parser/lexer is rudimentary, so it will not produce semantically meaningful error messages, just basic parse errors. Also it is kind of half automatic, half manual, which makes it much harder to include meaningful error context, as that context has to travel through 5+ functions and thus should be refected in all their signatures...

Let's not put a lot of effort into this right now, since we are not even sure for how long we will keep this lexer implementation. I think it might be worth to hack something that at least displays some basic context indicating where the error occurred though.

PR #91, opened 2025-10-28
feat: statically link cpp redist for MSVC target

vytautas-astrauskas-sensmetry:


andrius-puksta-sensmetry:
Why is this a draft? It passes CI, so presumably works well enough.

PR #97, opened 2025-11-03
feat(cli): warn that std libs are ignored if `--include-std` is not passed

### LINKED ISSUE #94: Generate warning or indication that standard library projects have been excluded from installation
We should consider printing a message in relevant commands (`install`, `add`, `sync`, etc) when the `--include-std` flag is *not* passed so that users are not surprised by standard library projects not being installed.

andrius-puksta-sensmetry:
Closes #94.

### DIFF
--- a/sysand/src/lib.rs
+++ b/sysand/src/lib.rs
@@ -4,7 +4,7 @@
 #[cfg(not(feature = "std"))]
 compile_error!("`std` feature is currently required to build `sysand`");
 
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 
 use anyhow::{Result, bail};
 
@@ -142,7 +142,14 @@ pub fn run_cli(args: cli::Args) -> Result<()> {
                     std::collections::HashMap::default()
                 };
 
-                command_sources_env(iri, version, !no_deps, current_environment, &provided_iris)
+                command_sources_env(
+                    iri,
+                    version,
+                    !no_deps,
+                    current_environment,
+                    &provided_iris,
+                    include_std,
+                )
             }
         },
         cli::Command::Lock { dependency_opts } => {
@@ -182,9 +189,10 @@ pub fn run_cli(args: cli::Args) -> Result<()> {
             };
 
             let provided_iris = if !include_std {
+                crate::logger::warn_std_deps();
                 known_std_libs()
             } else {
-                std::collections::HashMap::default()
+                HashMap::default()
             };
             let project_root = project_root.unwrap_or(std::env::current_dir()?);
             let lockfile = project_root.join(sysand_core::commands::lock::DEFAULT_LOCKFILE_NAME);
@@ -224,6 +232,7 @@ pub fn run_cli(args: cli::Args) -> Result<()> {
             } = dependency_opts;
             let index_base_urls = if no_index { None } else { Some(use_index) };
             let excluded_iris: HashSet<_> = if !include_std {
+                crate::logger::warn_std_deps();
                 known_std_libs().keys().cloned().collect()
             } else {
                 std::collections::HashSet::default()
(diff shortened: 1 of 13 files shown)


sysand/src/lib.rs
```diff
@@ -154,6 +161,10 @@ pub fn run_cli(args: cli::Args) -> Result<()> {
             let index_base_urls = if no_index { None } else { Some(use_index) };
 
             let provided_iris = if !include_std {
+                // TODO: lockfile will include std libs either way, but
+                // if `include_std` is passed, it will also include their sources
+                // Not sure if the warning is needed here.
+                crate::logger::warn_std_general();
```
andrius-puksta-sensmetry: Is the warning needed here? As far as I can tell, it does not meaningfully change anything. If `--include-std` is passed, std libs are resolved as all others, i.e. they include their kpar source. Otherwise, they are listed in the lock file by name.

sysand/tests/common/mod.rs
```diff
@@ -31,7 +30,7 @@ pub fn sysand_cmd_in<'a, I: IntoIterator<Item = &'a str>>(
     ]
     .concat();
     // NOTE had trouble getting test-temp-dir crate working, but would be better
-    let mut cmd = Command::cargo_bin("sysand")?;
+    let mut cmd = Command::new(assert_cmd::cargo::cargo_bin!("sysand"));
```
andrius-puksta-sensmetry: Old way was deprecated.

sysand/src/lib.rs
tilowiklundSensmetry: I think it's fine to skip the warning here.

sysand/src/logger.rs
```diff
@@ -32,3 +32,27 @@ fn format(buf: &mut Formatter, record: &Record<'_>) -> Result<(), std::io::Error
         }
     }
 }
+
+/// Print a warning that standard library package 'iri' should be provided by the tooling
+pub fn warn_std(iri: &str) {
+    let warn = "Warning";
+    let sp = " ";
+    let header = crate::style::WARN;
+    log::info!(
```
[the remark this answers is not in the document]
andrius-puksta-sensmetry: I wanted consistent styling. `log::warn!` looks out of place:
```plain
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2.60s
     Running `/Users/andrius/repos/sysand/target/debug/sysand add 'urn:kpar:systems-library' --include-std`
      Adding usage: 'urn:kpar:systems-library'
warning: SysML standard library packages will not be included in your sources,
         as they are assumed to be provided by your tooling.
         If you want to include them, use the '--include-std' flag
     Syncing env
```
I made it consistent:
```plain
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2.60s
     Running `/Users/andrius/repos/sysand/target/debug/sysand add 'urn:kpar:systems-library' --include-std`
      Adding usage: 'urn:kpar:systems-library'
     Warning SysML/KerML standard library package 'urn:kpar:systems-library'
             is ignored, as it should always be provided by your tooling.
             If you want to process it, pass the '--include-std' flag
     Syncing env
```

Not sure if it should stand out, but it's nice to keep consistency with other warnings. Will change to use `log::warn!`.

andrius-puksta-sensmetry:
> This looks good, but I think, at least for now, we need to always show some sort of warning when --include-std is not used.

This PR _does_ always print the warning, except for the lock command.

andrius-puksta-sensmetry:
It does:
```plain
$ sysand env install 'urn:kpar:systems-library'`
warning: SysML/KerML standard library package 'urn:kpar:systems-library'
         is ignored, as it should always be provided by your tooling.
         If you want to process it, pass the '--include-std' flag
     Syncing env
env is up to date
```
(it shouldn't sync the env if packages didn't change, will fix that)

tilowiklundSensmetry (approved):
:+1:

Issue #85, opened 2025-10-23
Lockfile uses absolute path for editable entries.


tilowiklundSensmetry:
In the current lock file implementation, editable entries get assigned absolute file system source paths. While for `file://`-URLs this may be expected (and might not be avoidable), it should not be the case for `editable` entries.

PR #100, opened 2025-11-06
Fix absolute path to current project in lockfile

victor-linroth-sensmetry:
Fixes the lockfiles path to current project from an absolute path to being simply `"."`.

### DIFF
--- a/sysand/src/commands/add.rs
+++ b/sysand/src/commands/add.rs
@@ -1,7 +1,7 @@
 // SPDX-FileCopyrightText: © 2025 Sysand contributors <opensource@sensmetry.com>
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
-use std::{collections::HashMap, sync::Arc};
+use std::{collections::HashMap, path::PathBuf, sync::Arc};
 
 use anyhow::Result;
 
@@ -46,7 +46,7 @@ pub fn command_add(
     if !no_lock {
         let index_base_urls = if no_index { None } else { Some(use_index) };
         crate::commands::lock::command_lock(
-            &project_root,
+            PathBuf::from("."),
             client.clone(),
             index_base_urls,
             &provided_iris,
--- a/sysand/src/lib.rs
+++ b/sysand/src/lib.rs
@@ -6,7 +6,7 @@ compile_error!("`std` feature is currently required to build `sysand`");
 
 use std::{
     collections::{HashMap, HashSet},
-    path::Path,
+    path::{Path, PathBuf},
     sync::Arc,
 };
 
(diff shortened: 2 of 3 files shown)


tilowiklundSensmetry:
Looks good, do we need to add an assertion that `.` is actually the correct path (so as to avoid accidentally making this assumption in the future, where these may be sub-paths of the project)?

victor-linroth-sensmetry:
> Looks good, do we need to add an assertion that `.` is actually the correct path (so as to avoid accidentally making this assumption in the future, where these may be sub-paths of the project)?

If we want to do something like workspaces later, I feel like `command_lock` will probably need quite a lot of reworking, and I'm not sure there would even be something like a "current project". In a way "current project" is `"."` by definition?

victor-linroth-sensmetry:
> In my min `.` is always some form of "working directory". I know @vytautas-astrauskas-sensmetry wants to push towards supporting some type of workspace functionality, where a workspace may contain multiple projects. If something like this happens, I think it not unlikely that lockfiles will exist at a workspace, rather than a project level. In this case `.` would probably be implicitly interpreted as the directory containing the workspace. In case additional functionality is implemented by someone other than us (or we simply forgot what we wrote here), I suspect this may lead to bugs.
> 
> I'd be totally fine with just `panic!`-ing here though, just to make the failure loud and clear and easy to debug.
> 
> EDIT: Specifically, we should panic if `.` is not the same as the directory `sync` would look in.

I don't really know what check for though. As far as I'm concerned `.` should always be interpreted as `project_root` (we may want to rename this if introducing workspaces, and we need to adapt project discovery to be a form of workspace discovery instead). Right now this is where we assume the current project to be. There's no way to assert that this is where the project is because we don't keep track of it any other way.

tilowiklundSensmetry:
Could we add an (integration) test case to ensure the correct editable entry and path is set?

tilowiklundSensmetry:
> Could we add an (integration) test case to ensure the correct editable entry and path is set?

Just found this too https://github.com/sensmetry/sysand/pull/102. I propose we get these two merged quickly and then we prioritise to have a good test suite for `lock` and `sync`.

andrius-puksta-sensmetry (approved):
LGTM

andrius-puksta-sensmetry:
Does this fix #85 or is there something else to do?

tilowiklundSensmetry:
> > Then why did we not catch the bug fixed by this MR?
> 
> Because it used to check for the absolute path. It's changed as part of this PR.

:facepalm: I'm either too tired or there is something funky with GitHub's interface, I was sure I checked the list of filenames

Issue #101, opened 2025-11-06
Support workspaces


vytautas-astrauskas-sensmetry:
We expect that at least some users will want to work on multiple projects simultaneously. For example, the SysML standard library is composed of [multiple projects](https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/tree/master/sysml.library), all of which are expected to have the same metadata matching a specific version of the specification. In package managers like Cargo, working on multiple projects is supported via the concept of a [workspace](https://doc.rust-lang.org/cargo/reference/workspaces.html). To avoid reinventing the wheel, we probably should just try to adapt an existing concept instead.

Requirements (the list is non-exhaustive and should be updated once we learn new things):

- [ ] All dependencies are considered together. This means:

    - [ ] A single lock file for all projects.
    - [ ] A single environment folder with dependencies for all projects.

- [ ] Commands should recognize that they are in a workspace and change their behavior accordingly.

Here is my proposal how each command should work when executed in the context of a workspace:

- [ ] `sysand init` and `sysand new`: should have a flag for automatically adding the created project to the workspace. If the flag is not submitted, they should report a warning that the workspace is detected and the user should manually update it.
- [ ] `sysand add`, `sysand remove`, `sysand include`, `sysand exclude`, `sysand info`, `sysand print-root`: when executed from a project's folder, should act normally. When executed from the workspace's folder, should show an error asking to use flag `--project <project name or one of its IRIs>` to specify in which project's context the command should be executed.
- [ ] `sysand build`: build kpars for all projects in the workspace and put them into the workspace output folder named `<project-name>-<project-version>.kpar`.
- [ ] `sysand lock`: should consider all projects in the workspace together and generate a single lock file.
- [ ] `sysand env`: should have a single environment for the workspace.
- [ ] `sysand env sources`: by default lists sources of all projects, but should have a flag for list of a single project.
- [ ] `sysand sync`: use the workspace's lock file and environment.
- [ ] `sysand sources`: list source files for all projects, unless a specific project is specified via flag.

I propose to identify the workspace by having `.workspace.json` file that lists the projects that belong to it. For example:

```json
{
    "projects": [
        {
            "path": "projectGroup1/project1",
            "iris": [
                "urn:kpar:project1"
            ]
        },
        {
            "path": "projectGroup1/project2",
            "iris": [
                "urn:kpar:project2"
            ]
        },
        {
            "path": "project3",
            "iris": [
                "urn:kpar:project3"
            ]
        }
[...]

PR #104, opened 2025-11-09
feat: add experimental workspace support for building kpars

vytautas-astrauskas-sensmetry:
Add experimental support for building kpars to make it easier to discuss the design at Reference Implementation WG. The design is not final and likely to change as mentioned in the warning printed each time the command is used.

Related issue: https://github.com/sensmetry/sysand/issues/101

### DIFF
--- a/core/src/commands/build.rs
+++ b/core/src/commands/build.rs
@@ -11,7 +11,10 @@ use crate::{
         local_src::LocalSrcError,
         utils::{FsIoError, ZipArchiveError},
     },
+    workspace::WorkspaceReadError,
 };
+#[cfg(feature = "filesystem")]
+use crate::{project::local_src::LocalSrcProject, workspace::Workspace};
 
 use super::include::IncludeError;
 
@@ -19,6 +22,8 @@ use super::include::IncludeError;
 pub enum KParBuildError<ProjectReadError> {
     #[error(transparent)]
     ProjectRead(ProjectReadError),
+    #[error(transparent)]
+    WorkspaceRead(#[from] WorkspaceReadError),
     #[error(transparent)]
     LocalSrc(#[from] LocalSrcError),
     #[error("incomplete project: {0}")]
@@ -39,6 +44,8 @@ pub enum KParBuildError<ProjectReadError> {
     Zip(#[from] ZipArchiveError),
     #[error("project serialization error: {0}: {1}")]
     Serialize(&'static str, serde_json::Error),
+    #[error("Internal error: {0}")]
+    InternalError(&'static str),
 }
 
 impl<ProjectReadError> From<FsIoError> for KParBuildError<ProjectReadError> {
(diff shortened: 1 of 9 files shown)


core/src/commands/build.rs
```diff
@@ -84,6 +91,20 @@ impl<ProjectReadError> From<IntoKparError<LocalSrcError>> for KParBuildError<Pro
...
+    project: &Pr,
+) -> Result<String, KParBuildError<Pr::Error>> {
+    let Some(project_info) = project.get_info().map_err(KParBuildError::ProjectRead)? else {
+        return Err(KParBuildError::MissingInfo);
+    };
+    Ok(format!(
+        "{}-{}.kpar",
+        project_info.name.replace(" ", "_").replace("-", "_"),
+        project_info.version
+    ))
```
andrius-puksta-sensmetry: The spec does not seem to constrain project's name in any way (it is noted as a `string` and no further limitations are given). It can thus in principle contain newlines, tabs and other junk. Therefore the normalization here should be more robust.

Though in practice the current way will work in the vast majority of cases.

sysand/tests/cli_build.rs
```diff
@@ -71,3 +71,92 @@ fn test_build() -> Result<(), Box<dyn std::error::Error>> {
...
+    {
+        let workspace_path = cwd.join("workspace.json");
+        let mut workspace_file = std::fs::File::create(workspace_path)?;
+        workspace_file.write_all(
+            b"{\"projects\": [\
+            {\"path\": \"subgroup/project1\", \"iris\": [\"urn:kpar:project1\"]},\
+            {\"path\": \"subgroup/project2\", \"iris\": [\"urn:kpar:project2\"]},\
+            {\"path\": \"project3\", \"iris\": [\"urn:kpar:project3\"]}\
+            ]}",
+        )?;
```
andrius-puksta-sensmetry: [Raw byte string literals](https://doc.rust-lang.org/reference/tokens.html#r-lex.token.str-byte-raw) don't need to escape `"`: `br#"{"projects":[{"path":"aaa"}]}"#`

core/src/discover.rs
```diff
@@ -7,19 +7,42 @@ use std::{
...
+    Some(LocalSrcProject { project_path: path })
+}
+
+pub fn current_workspace() -> Result<Option<Workspace>, io::Error> {
+    Ok(discover_workspace(current_dir()?))
+}
+
+pub fn discover_workspace<P: AsRef<Path>>(working_directory: P) -> Option<Workspace> {
+    let path = discover(working_directory, |path| {
+        path.join("workspace.json").is_file()
```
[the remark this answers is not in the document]
vytautas-astrauskas-sensmetry: Thanks for catching! Changed in 23b36f2bdce11a6c63216d648e16dd2460a79f07

andrius-puksta-sensmetry (approved):
Looking good.

PR #107, opened 2025-11-10
feat: add basic KerML support

vytautas-astrauskas-sensmetry:


core/src/symbols/mod.rs
```diff
@@ -545,6 +472,121 @@ pub fn top_level_sysml<R: io::Read>(mut reader: R) -> Result<Vec<String>, Extrac
...
+
+    let all = lex_source(&source)?;
+
+    let keywords = HashMap::from([
+        ("public", KeywordType::Simple),
+        ("protected", KeywordType::Simple),
+        ("private", KeywordType::Simple),
+        ("standard", KeywordType::Simple),
+        ("library", KeywordType::Simple),
+        ("package", KeywordType::Simple),
```
victor-linroth-sensmetry: Is there a reason for only considering packages for KerML while for SysML we check for everything? If I include a KerML file with a top level feature `sysand` just panics and crashes.

PR #116, opened 2025-11-13
feat: add release job for Java

vytautas-astrauskas-sensmetry:
This PR is a draft because it does not fit our current publishing workflow. Our other workflows are not doing releases on tagged commits. Instead, we promote a specific nightly release into a normal release. This approach does not work with Maven because of two reasons:

1. We need to choose whether we are building a snapshot or non-snapshot version before we start the build.
2. The easiest way to publish releases to Maven central is by using the corresponding plugin, which cannot be easily split into the phases of building the deployment bundle and uploading the bundle.


.github/workflows/java.yml
```diff
@@ -195,9 +215,71 @@ jobs:
...
+        gpg-passphrase: MAVEN_GPG_PASSPHRASE
     - name: Deploy Java bindings
       shell: bash
       run: python bindings/java/scripts/java-builder.py deploy
       env:
         MAVEN_USERNAME: ${{ secrets.MAVEN_CENTRAL_TOKEN_USERNAME }}
         MAVEN_PASSWORD: ${{ secrets.MAVEN_CENTRAL_TOKEN_PASSWORD }}
+        MAVEN_GPG_PASSPHRASE: ${{ secrets.MAVEN_CENTRAL_GPG_PASSPHRASE }}
+        # JAVA_BUILDER_RELEASE_JAR_VERSION: ${{ startsWith(github.ref, 'refs/tags/') }}
+        JAVA_BUILDER_RELEASE_JAR_VERSION: true
```
andrius-puksta-sensmetry: ~~newline~~ (for some reason gh showed missing newline, but not anymore)

.github/workflows/python.yml
andrius-puksta-sensmetry: ?

.github/workflows/test.yml
andrius-puksta-sensmetry: ?

PR #120, opened 2025-11-17
feat: make signing JARs optional

vytautas-astrauskas-sensmetry:


victor-linroth-sensmetry (approved):
Okay, seems to work now.

PR #121, opened 2025-11-17
feat: allow adding indexes in configuration files

victor-linroth-sensmetry:
Adds support for loading index from configuration files, and both indexes and configuration file can be given by environment variables.

Also changes a bit how extra indexes and configuration files are handled:

- Similar to uv there are regular (extra) indexes and default indexes, so command line options have changed to `--index` and `--default-index` instead of `--use-index`. Unlike uv multiple default indexes can be given in the command line (and environment variables) in order to better match the configuration files where there is no limit on number of default indexes.
- The `--config-file` option now adds a config file in addition to the automatically loaded ones.

See the added documentation for more details.

Also updated some minor formatting and lockfile names in the documentation and `README.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ A model interchange project is a collection of SysML or KerML files with
 additional metadata such as project name, versions, and the list of projects on
 which it depends. To create a new project called `my_project` run:
 
-```bash
+```console
 $ sysand new my_project
     Creating interchange project `my_project`
 ```
@@ -52,7 +52,7 @@ interchange project, consisting of two files `.project.json` and `.meta.json`.
 
 Inside the directory, we can ask for a basic overview of the project.
 
-```bash
+```console
 $ cd my_project
 $ sysand info
 Name: my_project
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -211,42 +211,48 @@ different project using `sysand`.
 ## Hosting a project index
 
 > [!important]
-> The structure of indices and `sysand_env` environments is still expected to
```
andrius-puksta-sensmetry: Both `indices` and `indexes` are correct.

README.md
```diff
@@ -211,42 +211,48 @@ different project using `sysand`.
...
 expose a `sysand_env` over HTTP.
 
 If you have an existing `sysand_env`, and you have a working Python 3 environment
 you can test this with
-```
+
+```sh
 $ python3 -m http.server -d sysand_env 8080
 Serving HTTP on 0.0.0.0 port 8080 (http://0.0.0.0:8080/) ...
 ```
```
andrius-puksta-sensmetry: `sh` is for shell commands. `$` line start and command output will not be correctly highlighted, as they will be interpreted as commands. I'd recommend using `plain` here

PR #124, opened 2025-11-21
docs: minor fixes

simonas-drauksas-sensmetry:
1. The lockfile was renamed in https://github.com/sensmetry/sysand/pull/58, this PR updates it in the user docs as well
2. The user guide had MacOS paths in one example output. Made it generic to fit with other example outputs


victor-linroth-sensmetry (approved):
LGTM

PR #130, opened 2025-11-28
feat: configuration of project sources

victor-linroth-sensmetry:
Main new features added by this PR:

- Config options for project sources in `sysand.toml` (or given config). Allows for portable specification of local and remote sources.
- Command line options `--from-path`, `--from-url`, `--as-editable`, `--as-local-src`, `--as-local-kpar`, `--as-remote-src`, `--as-remote-kpar` and `--as-remote-git` for `sysand add` that automatically updates config file. These also get removed with `sysand remove`.
- Documentation explaining the above.

In order to simplify implementing the above the `core` library was build out with the following changes:

- New macro crate with `ProjectRead` and `ProjectMut` derive macros for easy enum composition of project structs.
- New `CachedProject<Local, Remote>` struct for when the `Local` project is easier to access (e.g. on disk or in memory) vs `Remote`, but you still want `.sources()` to give the remote sources.
- Implementation of `CombinedResolver` simplified using `derive(ProjectRead)` and `CachedProject`.
- New `ProjectReference<Project>` struct. When working with `MemoryResolver` the projects need to be clonable, but this might not always be a good idea, e.g. some project hold ownership of temp dirs. `ProjectReference` is basically and `Arc` wrapper for a project to be used as a compatibility layer.
- New `AnyProject` struct combining the different projects supported for source configuration.
- `LocalSrcProject` and `LocalKparProject` now has optional nominal paths in addition to the regular path. This comes from the choice to have configured source always be relative to the project root but this not always being available everywhere. (Could perhaps be simplified if we added a Global Context).
- New `is_file` and `is_dir` functions added to `wrapfs` module that forwards errors when getting the metadata, unless that error is of kind `NotFound` in which case it returns `Ok(false)`.
- New `relativize_path` function for finding a relative path between to canonicalized paths with the same prefix.
- `do_sync` now supports Git sources.

In addition there where some small changes that came about when experimenting with different solutions:

- `MemoryEnv` can now be used with a generic project.
- `MemoryResolver` can be conveniently constructed from iterator/array/Vec.

These ended up not being used here, but they could be useful at some later point so where left in.

Lastly, the `verbose`/`quiet` options were removed from the config. They were mostly placeholders to begin with and now that there are more real options present in the config they don't serve much of a purpose. Removing them also means that the logger can be initialized almost immediately on start, opening up the potential for logging more things, like e.g. the loading of the config files.

### DIFF
--- a/sysand/src/commands/add.rs
+++ b/sysand/src/commands/add.rs
@@ -102,13 +102,13 @@ pub fn command_add<S: AsRef<str>, Policy: HTTPAuthentication>(
                     }
                 }
             }
-            ResolutionOutcome::UnsupportedIRIType(e) => bail!("unsupported URL: `{url}`\n{e}"),
+            ResolutionOutcome::UnsupportedIRIType(e) => bail!("unsupported URL `{url}`:\n{e}"),
             ResolutionOutcome::Unresolvable(e) => {
-                bail!("failed to resolve URL: `{url}`: {e}")
+                bail!("failed to resolve URL `{url}`:\n{e}")
             }
         }
         if source.is_none() {
-            bail!("unable to find project at URL: `{url}`")
+            bail!("unable to find project at URL `{url}`")
         }
         source
     } else if let Some(editable) = source_opts.as_editable {
--- a/docs/src/commands/add.md
+++ b/docs/src/commands/add.md
@@ -24,9 +24,9 @@ specify the path with the `--as-editable` flag.
 
 The affected configuration file will either be the one given with
 `--config-file` or (if `--no-config` is not present) the `sysand.toml` at the
-root of the project. If no configuration file is given and `--no-config` is set
-the usage will be added to the project but no source will be configured so
-future syncing will not take this into account.
+root of the project. If no configuration file is given and `--no-config` is
+set, the usage will be added to the project, but no source will be configured,
+so future syncing will not take this into account.
 
 ## Arguments
 


tilowiklundSensmetry:
Is there a particular reason for expanding `CombinedResolver` with an `Override` case, as opposed to having a generic `OverrideResolver<T>` and use `StandardResolver = OverrideResolver<CombinedResolver>`?

tilowiklundSensmetry:
Everything looks good to me in principle, would be a good to have a proper code review from @andrius-puksta-sensmetry before merging though.

core/src/project/local_kpar.rs
```diff
@@ -33,6 +33,7 @@ use super::utils::{FsIoError, ProjectDeserializationError, ToPathBuf, wrapfs};
 #[derive(Debug)]
 pub struct LocalKParProject {
     pub tmp_dir: Utf8TempDir,
+    pub nominal_path: Option<Utf8PathBuf>,
     pub archive_path: Utf8PathBuf,
     pub root: Option<Utf8PathBuf>,
```
andrius-puksta-sensmetry: Please document what all there paths are and how they are used. In particular `nominal_path`.

sysand/src/commands/env.rs
```diff
@@ -231,13 +272,16 @@ pub fn command_env_install_path<S: AsRef<str>, Policy: HTTPAuthentication>(
...
+            MemoryResolver::from(overrides),
             MemoryResolver {
                 iri_predicate: AcceptAll {},
                 projects: memory_projects,
             },
+        );
+        // TODO: Move out the runtime
+        let resolver = PriorityResolver::new(
+            override_resolver,
             standard_resolver(
```
andrius-puksta-sensmetry: Ok for now, but if we add one more layer of `PriorityResolver`, something will have to be done.

sysand/src/commands/add.rs
```diff
@@ -74,3 +132,59 @@ pub fn command_add<S: AsRef<str>, Policy: HTTPAuthentication>(
...
+        src_path.into()
+    };
+    Ok(src_path)
+}
+
+fn relativize(path: &Utf8Path, root: &Utf8Path) -> Utf8PathBuf {
+    // If prefixes (e.g. C: vs D: on Windows) differ, no relative path is possible.
+    if path.components().next() != root.components().next() {
+        return path.to_path_buf();
+    }
```
andrius-puksta-sensmetry: Not sure what to do about Windows UNC vs non-UNC paths, as they always have different first components. Relevant here is that std's `canonicalize` returns UNC paths on Windows, but calling that on both inputs here would be kind of wasteful.

sysand/src/commands/add.rs
```diff
@@ -74,3 +132,59 @@ pub fn command_add<S: AsRef<str>, Policy: HTTPAuthentication>(
...
+            );
+        }
+        path
+    } else {
+        src_path.into()
+    };
+    Ok(src_path)
+}
+
+fn relativize(path: &Utf8Path, root: &Utf8Path) -> Utf8PathBuf {
```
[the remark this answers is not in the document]
andrius-puksta-sensmetry: Also maybe return `Option<Utf8PathBuf>` so that caller can decide what to do in case relative path cannot be constructed.

core/src/project/cached.rs
```diff
@@ -0,0 +1,52 @@
...
+    fn read_source<P: AsRef<Utf8UnixPath>>(
+        &self,
+        path: P,
+    ) -> Result<Self::SourceReader<'_>, Self::Error> {
+        self.local.read_source(path)
+    }
+
+    fn sources(&self) -> Vec<crate::lock::Source> {
+        self.remote.sources()
+    }
```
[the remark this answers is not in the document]
victor-linroth-sensmetry: This is essentially just the same design as before (in `CombindedResolver`) but factored out into it's own struct. The cached project should return all the same data as the remote (hence no local sources) but with the fast access of the local. And yes that does mean it's on the user to make sure they represent the same underlying project. Maybe a better design is possibly but I decided to leave it as it was. I can add a docstring for clarification though.

docs/src/config/dependencies.md
```diff
@@ -0,0 +1,99 @@
...
+[`sysand add`](../commands/add.md) command.
+
+## Local projects
+
+To specify the source of a project that you have locally in a directory
+`./path/to/project` by the identifier `urn:kpar:my-project`, is done by adding
+the following entry to your `sysand.toml`.
+
+```toml
+[[project]]
```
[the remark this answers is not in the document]
victor-linroth-sensmetry: Yeah, I thought about that a too, but naming it something else breaks the symmetry with lockfile format. There is potentially more configurations we want to add here like e.g. aliasing (having multiple `identifiers`) and disabling checksums (in case the project is expected to change a lot like editable projects and sources pointing to git branches). I think having the lockfile and config mirror each other here would be nice.

sysand/src/commands/add.rs
```diff
@@ -74,3 +132,59 @@ pub fn command_add<S: AsRef<str>, Policy: HTTPAuthentication>(
...
+        }
+    }
+
+    let mut result = Utf8PathBuf::new();
+
+    for r in root_iter {
+        if let Utf8Component::Normal(_) = r {
+            result.push("..");
+        }
+    }
```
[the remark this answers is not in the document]
victor-linroth-sensmetry: > This is not always correct due to symlinks, unless both paths are already canonical.

Guess it's easiest to just canonicalize them both.

> This will also return an empty path if path and base are the same.

No it will be `.`? (Maybe some warning is appropriate here though since this doesn't make much sense for actual usage.)

> And the paths will use platform-dependent separator: \ on Windows and / everywhere? else. Is config.toml intended to be portable?

I'm pretty sure all paths get transformed to `Utf8UnixPaths` before going into any file.

sysand/src/commands/add.rs
```diff
@@ -74,3 +132,59 @@ pub fn command_add<S: AsRef<str>, Policy: HTTPAuthentication>(
...
+        if path.is_absolute() {
+            bail!(
+                "unable to find relative path from project root to `{}`",
+                path,
+            );
+        }
+        path
+    } else {
+        src_path.into()
+    };
```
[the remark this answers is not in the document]
victor-linroth-sensmetry: Added a check.

core/src/project/utils.rs
```diff
@@ -203,6 +204,58 @@ pub mod wrapfs {
...
+    ///   other than the path not being found (e.g., permission denied).
+    ///
+    /// # Errors
+    ///
+    /// Returns an [`FsIoError`] if metadata retrieval fails for any reason
+    /// other than [`std::io::ErrorKind::NotFound`].
+    pub fn is_file<P: AsRef<Utf8Path>>(path: P) -> Result<bool, Box<FsIoError>> {
+        match metadata(path) {
+            Ok(metadata) => Ok(metadata.is_file()),
+            Err(err) if matches!(err.as_ref(), FsIoError::Metadata(_, e) if e.kind() == ErrorKind::NotFound) => {
```
andrius-puksta-sensmetry: `std::fs::metadata` should be used here

core/src/project/utils.rs
```diff
@@ -250,3 +303,231 @@ pub enum ZipArchiveError {
...
+        let root = if cfg!(windows) {
+            Utf8Path::new(r"C:\d\e\f")
+        } else {
+            Utf8Path::new("/d/e/f")
+        };
+        let relative = if cfg!(windows) {
+            Utf8Path::new(r"..\..\..\a\b\c")
+        } else {
+            Utf8Path::new("../../../a/b/c")
+        };
```
andrius-puksta-sensmetry: This also

sysand/src/commands/add.rs
```diff
@@ -23,14 +36,121 @@ pub fn command_add<S: AsRef<str>, Policy: HTTPAuthentication>(
...
+            add_project_source_to_config(&path, &iri, &source)?;
+
+            config.projects.push(ConfigProject {
+                identifiers: vec![iri.as_ref().to_string()],
+                sources: vec![source],
+            });
+        } else {
+            bail!("must provide config file for specifying project source")
+        }
+    }
```
andrius-puksta-sensmetry: This is inconsistent with docs, as they state that not providing a config is allowed.

sysand/src/commands/add.rs
```diff
@@ -74,3 +195,19 @@ pub fn command_add<S: AsRef<str>, Policy: HTTPAuthentication>(
 
     Ok(())
 }
+
+fn get_relative<P: Into<Utf8PathBuf> + AsRef<Utf8Path>>(
```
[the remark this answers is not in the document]
andrius-puksta-sensmetry: But if `root_path` is canonical, then `current_dir` will likely not match on Windows...

core/src/commands/lock.rs
```diff
@@ -157,17 +163,29 @@ pub type EditableLocalSrcProject = EditableProject<LocalSrcProject>;
...
         LocalSrcProject {
-            project_path: path.to_path_buf(),
+            nominal_path: Some(path.to_path_buf()),
+            project_path: project_root
+                .as_ref()
+                .join(path.as_ref())
+                .canonicalize_utf8()
+                .map_err(|e| {
+                    LockError::Io(
+                        FsIoError::Canonicalize(project_root.to_path_buf().join(path.as_ref()), e)
```
andrius-puksta-sensmetry:
```suggestion
            project_path: project_root
                .as_ref()
                .join(path)
                .canonicalize_utf8()
                .map_err(|e| {
                    LockError::Io(
                        FsIoError::Canonicalize(project_root.as_ref().join(path), e)
```

sysand/src/cli.rs
```diff
@@ -1270,6 +1272,44 @@ pub struct ResolutionOptions {
...
+    /// Add usage as a local interchange project at PATH and
+    /// update configuration file attempting to guess the
+    /// source from the PATH
+    #[arg(long, value_name = "PATH", group = "source")]
+    pub from_path: Option<String>,
+    /// Add usage as a remote interchange project at URL and
+    /// update configuration file attempting to guess the
+    /// source from the URL
+    #[arg(long, value_name = "URL", group = "source")]
+    pub from_url: Option<String>,
```
```suggestion
    pub from_url: Option<Iri<String>>,
```
Same for all other IRIs/URLs here.

core/src/lock.rs
```diff
@@ -270,9 +274,13 @@ impl Lock {
             let version = Version::parse(&project.version)
                 .inspect_err(|err| {
                     log::warn!(
-                        "invalid semantic version '{}' for project '{:?}'\n{}",
+                        "invalid semantic version '{}' for project {}\n{}",
```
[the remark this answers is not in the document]
victor-linroth-sensmetry: Same

core/src/resolve/memory.rs
```diff
@@ -13,14 +13,41 @@ use crate::{
 #[derive(Debug)]
 pub struct MemoryResolver<Predicate, ProjectStorage: Clone> {
     pub iri_predicate: Predicate,
-    pub projects: HashMap<fluent_uri::Iri<String>, Vec<ProjectStorage>>,
+    pub projects: HashMap<Iri<String>, Vec<ProjectStorage>>,
```
andrius-puksta-sensmetry: Is the `Vec` here for multiple versions of the same IRI?

sysand/src/commands/add.rs
```diff
@@ -23,18 +36,130 @@ pub fn command_add<S: AsRef<str>, Policy: HTTPAuthentication>(
...
+                        }
+                        Err(err) => {
+                            log::debug!("skipping candidate project: {err}");
+                        }
+                    }
+                }
+            }
+            ResolutionOutcome::UnsupportedIRIType(e) => bail!("unsupported URL: `{url}`\n{e}"),
+            ResolutionOutcome::Unresolvable(e) => {
+                bail!("failed to resolve URL: `{url}`: {e}")
```
andrius-puksta-sensmetry:
```suggestion
                bail!("failed to resolve URL `{url}`: {e}")
```

sysand/src/commands/add.rs
```diff
@@ -23,18 +36,130 @@ pub fn command_add<S: AsRef<str>, Policy: HTTPAuthentication>(
...
+            }
+        }
+        if source.is_none() {
+            bail!("unable to find project at URL: `{url}`")
```
andrius-puksta-sensmetry:
```suggestion
            bail!("unable to find project at URL `{url}`")
```

PR #144, opened 2025-12-19
fix(docs): update docs to match 0.0.7 release

victor-linroth-sensmetry:
Updates "Dependency options" to " Resolution options" and the descriptions to match CLI.

Also fixes markdown in version constraint argument for add command.

### DIFF
--- a/docs/src/commands/build.md
+++ b/docs/src/commands/build.md
@@ -1,6 +1,7 @@
 # `sysand build`
 
-Build a KerML Project Archive (KPAR)
+Build a KerML Project Archive (KPAR). If executed in a workspace outside of a
+project, builds all projects in the workspace.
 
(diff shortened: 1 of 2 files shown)


andrius-puksta-sensmetry:
There are a bunch more places where docs are out of date: 

- `clone` command is not mentioned (forgot to add it)
- many other commands had some of their docstrings updated
- `info` command page should be expanded to include all its subcommands and their docs

Basically the docs should IMO contain at least (equivalent to) all the help output from all commands and subcommands.

victor-linroth-sensmetry:
Adding documentation for `info` and `clone` can be separate PR:s. This is mainly intended to patch the changes to existing stuff. I can maybe try and compare to see if I find some more things.

PR #151, opened 2025-12-31
refactor: use camino's UTF-8 paths everywhere

### LINKED ISSUE #136: Require all paths to be UTF-8
Sysand currently handles paths inconsistently. Sometimes they are passed around as `String` (as they are less annoying to deal with, e.g. strings implement `Display`), sometimes as `Path`. There are a bunch of places where conversion from `Path` to `String` happens and each place has to somehow report possible conversion errors, which is annoying and tedious. This inconsistent usage also means that we don't get the main benefit of using a `Path` - support for invalid UTF-8 paths.

I propose we use [camino crate](https://docs.rs/camino/latest/camino/index.html) to always have UTF-8 paths and enforce them at API boundaries where possible and directly require them in CLI. camino's Utf8Path(Buf) paths can be converted to/from string/str for free with no checking/reallocation required. camino also wraps std's path and file APIs to consume and return only UTF-8 paths. They also conveniently implement `AsRef<str>` and `AsRef<Path>`, so can be used to call most std file I/O functions.

andrius-puksta-sensmetry:
Also use `camino-tempfile` instead of `tempfile` to get UTF-8 paths for tempfiles.

Closes #136.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -3,7 +3,7 @@
 #
 #    maturin generate-ci github --pytest --manifest-path bindings/py/Cargo.toml
 #
-name: Build/test/publish python packages
+name: Python
 
 on:
   push:


core/src/discover.rs
```diff
@@ -36,36 +36,36 @@ pub fn discover_workspace<P: AsRef<Path>>(working_directory: P) -> Option<Worksp
...
 
 // TODO: Improve the logic here, this is probably too simple
-fn discover<P: AsRef<Path>, F: Fn(&Path) -> bool>(
+fn discover<P: AsRef<Utf8Path>, F: Fn(&Utf8Path) -> bool>(
     working_directory: P,
     predicate: F,
-) -> Option<PathBuf> {
-    let mut current: PathBuf = working_directory.as_ref().to_path_buf();
+) -> Option<Utf8PathBuf> {
+    let mut current = working_directory.as_ref().to_path_buf();
```
victor-linroth-sensmetry:
```suggestion
    let mut current = working_directory.as_ref().to_owned();
```
Maybe a bit more idiomatic. Same for line 53 and 69.

victor-linroth-sensmetry (commented):
Question: Should `wrapfs` also be converted to camino paths/pathbufs? If there is nothing hindering us from doing this it seems like it could make things a bit more concise and uniform.

core/src/env/local_directory.rs
```diff
@@ -32,28 +32,33 @@ pub const ENTRIES_PATH: &str = "entries.txt";
...
         .into_iter()
         .filter_map(|e| e.ok())
-        .filter(|e| e.file_type().is_dir())
-        .map(|e| e.into_path())
+        .filter_map(|e| {
+            if e.file_type().is_dir() {
+                Utf8PathBuf::from_path_buf(e.into_path()).ok()
+            } else {
+                None
+            }
```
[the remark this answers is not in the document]
andrius-puksta-sensmetry: Done in [dd1cceb](https://github.com/sensmetry/sysand/pull/151/commits/dd1ccebd441098d73a556dc4a4dedadb117e7a8e)

core/src/resolve/file.rs
```diff
@@ -51,11 +56,19 @@ pub const SCHEME_FILE: &Scheme = Scheme::new_or_panic("file");
 
 /// Try to obtain a file path from `uri` with `file` scheme. If path
 /// is present, it is always absolute according to URI spec
-fn try_file_uri_to_path(uri: &fluent_uri::Iri<String>) -> Option<PathBuf> {
+fn try_file_uri_to_path(
+    uri: &fluent_uri::Iri<String>,
+) -> Option<Result<Utf8PathBuf, FileResolverError>> {
```
victor-linroth-sensmetry: Having the return type be `Result<Option<Utf8PathBuf>, FileResolverError>` seems like the more common way of doing things. Then you can just `match try_file_uri_to_path(uri)?` on line 126 and not do the nesting.

PR #154, opened 2026-01-07
feat: allow adding/removing usages by path

### LINKED ISSUE #206: `sysand add urn:kpar:sysmod` broken
Testing https://github.com/sensmetry/sysand/pull/183 i observed that access of index was already broken in main, where I test things like this:

```sh
d=$(mktemp -d) && cd "$d" \
  && cargo run --manifest-path ~/dev/sensmetry/sysand/Cargo.toml \
    init \
  && cargo run --manifest-path ~/dev/sensmetry/sysand/Cargo.toml \
    add urn:kpar:sysmod
```

The failure stems from https://github.com/sensmetry/sysand/pull/130, verified by testing before and after.

I used codex (GPT 5.4) to debug and review the situation, and it came up with some changes https://github.com/sensmetry/sysand/pull/205.

andrius-puksta-sensmetry:
Users want to do this anyway (at least for experimenting), so might as well accomodate them.

## Changes

- Add `--path` argument to `add` and `remove` that allows adding/removing path usages by converting them to `file://` URL. Recommend using workspaces instead of paths.
- Guide users to use path usage (both for add and remove) if IRI parsing fails
- Fail when building a kpar if path usages are present, unless `--allow-path-usage` is passed (in that case a warning is still produced)
- Guide users to use `init` instead of `new` (`sysand new` was mentioned in AoSysML, so this at least gives users something actionable)
- Add docs about supported IRIs
- (unrelated) Add workspace docs

## Fixes

- closes #206; added an integration test for it

### DIFF
--- a/docs/src/commands/build.md
+++ b/docs/src/commands/build.md
@@ -34,6 +34,7 @@ if none is found uses the current directory instead.
   - `deflated`: Compress the files using Deflate
 
   [default: `deflated`]
+
 - `-a`, `--allow-path-usage` Allow usages of local paths (`file://`).
   Warning: using this makes the project not portable between different
   computers, as `file://` URL always contains an absolute path.


tilowiklundSensmetry:
> So while using file urls for quick experimentation is fine, it's non-portable so generally bad practise and should not be encouraged. As such I would prefer not to have a nice and simple flag like `--path` for this, and instead have something with a bit more friction to it.
> 
> If we are to use a `--path` flag it should be part of the more general configuration of sources I've been working on (I've not committed to a UI for that though and will want some feedback before finalising).

I do think we need something that is (equivalently) easy to use for learning purposes, and `--path` is sort of the only thing I can think of.

A complex alternative option would be to have the Sysand client maintain some sort of "local index" that the user can either install, or register projects to. This is sort of doable already (using `env install`), but would need to be made a bit more ergonomic.

consideRatio:
There was a mistaken "request changes here", so i dismissed it by re-requesting review from @victor-linroth-sensmetry.

@andrius-puksta-sensmetry let me know if you need help getting this to land in one way or another. It would be great to reduce the open PRs further!

PR #165, opened 2026-01-26
feat(cli): rudimentary support for http authentication

tilowiklundSensmetry:
This MR implements the basic infrastructure for adding authentication to HTTP requests (to indices, project source, ...). It is an intermediate MR to https://github.com/sensmetry/sysand/pull/157, which intends to bring more complete support, both in terms of supported authentication schemes and configurability.

This MR only implements:
- Basic authentication scheme (username:password)
- Configurable in the CLI, via environment variables

The CLI looks for triplets of environment variables following the pattern `SYSAND_CRED_<X>`, `SYSAND_CRED_<X>_BASIC_USER`, `SYSAND_CRED_<X>_BASIC_PASS`. The `<X>` part is arbitrary, but every `<X>` has to appear either not at all, or for all three patterns. The first variable is a glob pattern to match URLs to allow authentication for, while the other two provide the actual credentials. 

Example
```
SYSAND_CRED_FOO='https://*.foo.com/**' SYSAND_CRED_FOO_BASIC_USER="bar" SYSAND_CRED_BASIC_PASS="baz" sysand <OP>
```
Would allow the use of `bar:baz` as credentials for urls such as `https://www.foo.com/a/b/c`, `https://index.foo.com/hey.kpar`, ...

Credentials are only actually sent if an initial request generates a 4xx status. The strictly correct behaviour here would be to try sending credentials only in response to an explicit `401` with `WWW-Authenticate` header, but this would, I believe, be incompatible with using, for example, private GitLab/GitHub pages.

### DIFF
--- a/core/src/resolve/standard.rs
+++ b/core/src/resolve/standard.rs
@@ -22,7 +22,8 @@ use reqwest_middleware::ClientWithMiddleware;
 
 pub type LocalEnvResolver = EnvResolver<LocalDirectoryEnvironment>;
 
-pub type RemoteIndexResolver<Policy> = SequentialResolver<EnvResolver<HTTPEnvironmentAsync<Policy>>>;
+pub type RemoteIndexResolver<Policy> =
+    SequentialResolver<EnvResolver<HTTPEnvironmentAsync<Policy>>>;
 
 type StandardResolverInner<Policy> = CombinedResolver<
     FileResolver,


docs/src/SUMMARY.md
```diff
@@ -4,6 +4,7 @@
   - [Installation](getting_started/installation.md)
   - [Tutorial](getting_started/tutorial.md)
 - [Project metadata](metadata.md)
+- [Auhtentication](authentication.md)
 - [Commands](commands.md)
   - [sysand init](commands/init.md)
   - [sysand new](commands/new.md)
```
andrius-puksta-sensmetry: (unrelated) remove `sysand new` line here.

sysand/tests/common/mod.rs
```diff
@@ -21,10 +22,11 @@ pub fn fixture_path(name: &str) -> Utf8PathBuf {
     path
 }
 
-pub fn sysand_cmd_in<'a, I: IntoIterator<Item = &'a str>>(
+pub fn sysand_cmd_in_with<'a, I: IntoIterator<Item = &'a str>>(
     cwd: &Utf8Path,
     args: I,
     cfg: Option<&str>,
+    env: &IndexMap<String, String>,
```
andrius-puksta-sensmetry: Maybe change this (and other methods here) to take `IndexMap<AsRef<str>, AsRef<str>>` to avoid `.to_string()`/`.to_owned()` everywhere?

docs/src/SUMMARY.md
tilowiklundSensmetry: I'll start a new branch/MR for documentation cleanup.

docs/src/authentication.md
```diff
@@ -0,0 +1,49 @@
...
+https://projects.example.com/projects/myproject/versions.txt
+```
+
+In the wildcard pattern, `?` matches any single letter, `*` matches any sequence of characters
+not containing `/`, and `**` matches any sequence of characters possibly including `/`.
+
+Credentials will *only* be sent to URLs matching the pattern, and even then only if an 
+unauthenticated response produces a status in the 4xx range. If multiple patterns match, they will
+be tried in an arbitrary order, after the initial unauthenticated attempt, until one results in a
+response not in the 4xx range.
```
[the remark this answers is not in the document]
tilowiklundSensmetry: > What use case do you have in mind where it is not a logical error to have overlapping URL patterns?

I'm mostly thinking of situations where a user has low security account for a general set of resources and a high security account for a limited (typically subset) of resources. Manually disentangling the corresponding glob patterns is a hassle. 

If we make it an error (and maybe introduce an optional priority number or something like that), I think it should be "statically" detected at startup. I think this is possible to determine, but not trivial to implement.

core/src/env/reqwest_http.rs
```diff
@@ -26,11 +29,12 @@ use crate::{
 
 use futures::{AsyncBufReadExt as _, StreamExt as _};
 
-pub type HTTPEnvironment = AsSyncEnvironmentTokio<HTTPEnvironmentAsync>;
+pub type HTTPEnvironment = AsSyncEnvironmentTokio<HTTPEnvironmentAsync<StandardHTTPAuthentication>>;
 
 #[derive(Debug)]
-pub struct HTTPEnvironmentAsync {
+pub struct HTTPEnvironmentAsync<Pol> {
```
victor-linroth-sensmetry: Readability: Can we just have this be `Policy`? Surely no one is working on screen so small that three characters makes a difference?

core/src/config/mod.rs
```diff
@@ -133,6 +144,7 @@ mod tests {
                 url: "http://www.example.com".to_string(),
                 ..Default::default()
             }]),
+            auth: None,
```
victor-linroth-sensmetry: Put in an actual value here to test that merging with default works.

docs/src/authentication.md
```diff
@@ -0,0 +1,49 @@
...
+
+- HTTP(S) with digest access, (fixed) bearer token, and OAuth2 device authentication
+- Git with private-key and basic access authentication
+
+## Configuring
+
+At the time of writing authentication can only be configured through environment variables.
+Providing credentials is done by setting environment variables following the pattern
+
+```
```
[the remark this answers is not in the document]
andrius-puksta-sensmetry: Note: use `text` instead of `plain`, as only `text` seems to be [explicitly supported by mdbook](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#attributes). `text` also supported by both [rustdoc](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#attributes) and GitHub.

core/src/project/reqwest_src.rs
```diff
@@ -180,7 +185,19 @@ impl ProjectReadAsync for ReqwestSrcProjectAsync {
     }
 
     async fn is_definitely_invalid_async(&self) -> bool {
-        match join!(self.head_info().send(), self.head_meta().send()) {
+        let info_url = self.info_url();
+        let info_request = move |client: &ClientWithMiddleware| client.head(info_url.clone());
+        let info_resp = self
+            .auth_policy
+            .with_authentication(&self.client, &info_request);
```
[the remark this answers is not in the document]
tilowiklundSensmetry: Considering there are not many places we actually make requests I won't change this design for now, unless @victor-linroth-sensmetry @andrius-puksta-sensmetry either of you have a better idea.

core/src/project/reqwest_kpar_download.rs
```diff
@@ -27,10 +32,11 @@ use super::utils::{FsIoError, wrapfs};
 /// Downloads the full archive to a temporary directory and then accesses it using
 /// `LocalKParProject`.
 #[derive(Debug)]
-pub struct ReqwestKparDownloadedProject {
+pub struct ReqwestKparDownloadedProject<Pol> {
```
victor-linroth-sensmetry:
```suggestion
pub struct ReqwestKparDownloadedProject<Policy> {
```

core/src/project/reqwest_src.rs
```diff
@@ -27,14 +29,15 @@ use crate::{
 /// is accessed by
 /// GET https://www.example.com/project/M%C3%ABkan%C3%AFk/K%C3%B6mmand%C3%B6h.sysml
 #[derive(Clone, Debug)]
-pub struct ReqwestSrcProjectAsync {
+pub struct ReqwestSrcProjectAsync<Pol> {
```
victor-linroth-sensmetry:
```suggestion
pub struct ReqwestSrcProjectAsync<Policy> {
```

core/src/project/reqwest_src.rs
```diff
@@ -27,14 +29,15 @@ use crate::{
...
     /// (reqwest) HTTP client to use for GET requests
     pub client: reqwest_middleware::ClientWithMiddleware, // Internally an Arc
     /// Base-url of the project
     pub url: reqwest::Url,
+    pub auth_policy: Arc<Pol>,
 }
 
-impl ReqwestSrcProjectAsync {
+impl<Pol> ReqwestSrcProjectAsync<Pol> {
```
victor-linroth-sensmetry:
```suggestion
impl<Policy> ReqwestSrcProjectAsync<Policy> {
```

core/src/resolve/standard.rs
```diff
@@ -21,29 +22,29 @@ use reqwest_middleware::ClientWithMiddleware;
 
 pub type LocalEnvResolver = EnvResolver<LocalDirectoryEnvironment>;
 
-pub type RemoteIndexResolver = SequentialResolver<EnvResolver<HTTPEnvironmentAsync>>;
+pub type RemoteIndexResolver<Pol> = SequentialResolver<EnvResolver<HTTPEnvironmentAsync<Pol>>>;
 
-type StandardResolverInner = CombinedResolver<
+type StandardResolverInner<Pol> = CombinedResolver<
```
victor-linroth-sensmetry:
```suggestion
type StandardResolverInner<Policy> = CombinedResolver<
```

PR #166, opened 2026-01-30
Consolidate docs into docbook

andrius-puksta-sensmetry:
Source(s) -> destination:

- `README.md` + https://beta.sysand.org/user_guide.html + `docs/src/getting_started/tutorial.md` -> `docs/src/getting_started/tutorial.md`
- https://beta.sysand.org/publishing_guide.html -> `docs/src/publishing.md`
- `README.md` + https://beta.sysand.org/deployment_guide.html -> `docs/src/hosting_index.md`
- https://docs.sensmetry.com/modeler/cli.html#install-cli + `README.md` + https://beta.sysand.org/download.html + `docs/src/getting_started/installation.md` -> `docs/src/getting_started/installation.md`

### DIFF
--- a/docs/src/getting_started/installation.md
+++ b/docs/src/getting_started/installation.md
@@ -78,12 +78,12 @@ After downloading the appropriate file, installation depends on your platform:
 
 It is recommended to then [verify the installation](#verify-the-installation).
 
-## Windows
+### Windows
 
 The downloaded binary can either be installed manually or by running
 a few PowerShell commands.
 
-### Manual installation
+#### Manual installation
 
 1. Move the downloaded `.exe` file to `%LOCALAPPDATA%\Programs\Sysand\sysand.exe`
 2. Add to `PATH` via Environment Variables:
@@ -93,7 +93,7 @@ a few PowerShell commands.
     3. Click "New" and add `%LOCALAPPDATA%\Programs\Sysand`
     4. Click "Ok" to save
 
-### PowerShell installation
+#### PowerShell installation
 
 1. Open PowerShell
 2. Run these commands:
@@ -127,9 +127,9 @@ if ($currentPath -notlike "*$newPath*") {
 > Restart your terminal after installation for `PATH` changes to take effect.
 
 
-## macOS
+### macOS
 
-### System installation (requires `sudo`)
+#### System installation (requires `sudo`)
 
 1. Open Terminal
 2. Make the binary executable and move to a folder in `PATH` by running the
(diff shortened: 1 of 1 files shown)


docs/src/getting_started/tutorial.md
```diff
@@ -24,151 +35,160 @@ Version: 0.0.1
...
+The latter directory stores the added project and its usages.
 
 ```console
 $ sysand info
 Name: my_project
 Version: 0.0.1
 Usages:
-    https://www.omg.org/spec/KerML/20250201/Semantic-Library.kpar
+    urn:kpar:sysmod
+    TODO: add second usage from URL here
```
simonas-drauksas-sensmetry: > TODO: add second usage from URL here

If we use the std lib kpar URL above, then `sysand info` will not show the URL here, only `sysand info --include-std`

docs/src/getting_started/installation.md
```diff
@@ -1,12 +1,199 @@
...
+
+You should see an output similar to: `sysand X.Y.Z` 
+
+
+## PyPI
+
+Sysand is [published on PyPI](https://pypi.org/project/sysand/) and includes both
+the CLI and Sysand Python library.
+
+It can be installed using [pip](https://pip.pypa.io/en/stable/). Run the following
```
victor-linroth-sensmetry:
```suggestion
It can be installed using [pip](https://pip.pypa.io/en/stable/installation/). Run the following
```

- Pip also installs as `pip3` on some platforms.
- Isn't it generally recommended to only do `pip install` in virtual environments?
- Maybe add instructions for installation using `pipx` and `uv tool` for those that want it available outside of virtual environments.
- Should we move the PyPi installation to the top as the preferred option for the non-technical user?

docs/src/getting_started/tutorial.md
simonas-drauksas-sensmetry: Let's take this example out for now, due to https://github.com/sensmetry/sysand/pull/166#discussion_r2757685932

docs/src/getting_started/installation.md
```diff
@@ -1,12 +1,237 @@
...
+<!-- markdownlint-enable MD033 MD013 -->
+After downloading the appropriate file, installation depends on your platform:
+
+- [Windows (both x86_64 and ARM64)](#windows)
+- [macOS (both Intel and ARM64 (a.k.a. Apple Silicon))](#macos)
+- [Linux (both x86_64 and ARM64)](#linux)
+
+It is recommended to then [verify the installation](#verify-the-installation).
+
+## Windows
```
simonas-drauksas-sensmetry:
```suggestion
### Windows
```

This (and all other platforms and their subsections) should be pushed down one heading level, since the "Download latest release" section is level 2, and these subsections are only relevant for that.

PR #175, opened 2026-02-11
feat: add environment metadata file `sysand_env/env.toml`

victor-linroth-sensmetry:
This PR adds a `env.toml` containing metadata for the projects installed in the local environment
(inluding editable projects not present inside the `sysand_env` folder).

Each project has the following fields:

- `publisher`: String (optional).
  Publisher of the project. Intended for display purposes.
- `name`: String (optional).
  Name of the project. Intended for display purposes.
- `version`: String (required).
  Version of the project.
- `path`: String (required).
  Path to the root directory of the project.
  If the project is not `editable` this should be relative
  to the env directory and otherwise it should be relative
  to the workspace root.
- `identifiers`: Array of strings (required for non-editable projects)
  List of identifiers (IRIs) used for the project.
  The first identifier is to. be considered the canonical
  identifier, and if the project is not `editable` this
  is the IRI it is installed as. The rest are considered
  as aliases. Can only be empty for `editable` projects.
- `usages`: Array of strings.
  Usages of the project. Intended for tools needing to
  track the interdependence of project in the environment.
- `editable`: bool.
  Indicator of wether the project is fully installed in
  the environment or located elsewhere.
-  `workspace`: Indicator of wether the project is part of a workspace.

The current implementation doesn't change how `projects.txt` works and is intended to offer a transition step between env structures. The `env.toml` isn't directly managed by
`LocalDirectoryEnvironment` since such a thing will likely require a change in the
`WriteEnvironment` trait. In particular it would have to offer the ability to give alias identifiers in addition to the main IRI used for installation (since we want to communicate the dependency structure between projects we need all aliases as well, at least if we want to incrementally add or remove from the environment).

Also since the current implementation of workspaces only affects the `build` command, the `workspace` field isn't really effective at the moment, but should work as expected once the workspace feature catches up.

Example:

```toml
# This file is automatically generated by Sysand and is not intended to be edited manually.

version = "0.1"

[[project]]
name = "SYSMOD"
identifiers = [
    "urn:kpar:sysmod",
]
version = "5.0.0-alpha.2"
path = "a0aacee34dd4cd5e2d07ab43d5e30772ec86dbf3c8fafb033bad338dd7a0f02e/5.0.0-alpha.2.kpar"
usages = [
    "urn:kpar:analysis-library",
    "urn:kpar:cause-and-effect-library",
    "urn:kpar:data-type-library",
    "urn:kpar:function-library",
    "urn:kpar:geometry-library",
    "urn:kpar:metadata-library",
    "urn:kpar:quantities-and-units-library",
    "urn:kpar:requirement-derivation-library",
    "urn:kpar:semantic-library",
    "urn:kpar:systems-library",
]

[[project]]
name = "Testing"
version = "0.0.1"
path = "."
usages = [
    "urn:kpar:sysmod",
]
editable = true
files = [
    "test.sysml",
    "test.kerml",
]

```

### DIFF
--- a/core/src/env/local_directory/metadata.rs
+++ b/core/src/env/local_directory/metadata.rs
@@ -90,7 +90,7 @@ impl Lock {
 #[derive(Debug, Deserialize)]
 pub struct EnvMetadata {
     pub version: String,
-    #[serde(rename = "project", skip_serializing_if = "Vec::is_empty", default)]
+    #[serde(rename = "project", default)]
     pub projects: Vec<EnvProject>,
 }
 
@@ -176,7 +176,7 @@ impl EnvMetadata {
 
     pub fn add_project(&mut self, project: EnvProject) {
         if let Some(found) = self.find_project(&project.identifiers, &project.version) {
-            self.projects[found].merge(&project);
+            self.projects[found].merge_identifiers(&project);
         } else {
             self.projects.push(project);
         }
@@ -249,22 +249,22 @@ pub struct EnvProject {
     /// identifier, and if the project is not `editable` this
     /// is the IRI it is installed as. The rest are considered
     /// as aliases. Can only be empty for `editable` projects.
-    #[serde(skip_serializing_if = "Vec::is_empty", default)]
+    #[serde(default)]
     pub identifiers: Vec<String>,
     /// Usages of the project. Intended for tools needing to
     /// track the interdependence of project in the environment.
-    #[serde(skip_serializing_if = "Vec::is_empty", default)]
+    #[serde(default)]
     pub usages: Vec<String>,
     /// Indicator of wether the project is fully installed in
     /// the environment or located elsewhere.
-    #[serde(skip_serializing_if = "bool::is_false", default)]
+    #[serde(default)]
     pub editable: bool,
     /// In case of an `editable` project these are the files
     /// belonging to the project. Intended for tools that
     /// are not able to natively parse and understand the
     /// projects `.meta.json` file. Paths should be relative
     /// to the `path` of the project.
-    #[serde(skip_serializing_if = "bool::is_false", default)]
+    #[serde(default)]
     pub workspace: bool,
 }
 
@@ -291,6 +291,9 @@ impl EnvProject {
         if self.editable {
             table.insert("editable", value(true));
         }
+        if self.workspace {
+            table.insert("workspace", value(true));
+        }
 
         table
     }
@@ -298,10 +301,10 @@ impl EnvProject {
     /// Adds identifiers from other project.
     /// Should only be done if the underlying projects are the same.
     /// In particular they must have the same version.
-    pub fn merge(&mut self, other: &EnvProject) {
+    pub fn merge_identifiers(&mut self, other: &EnvProject) {
         assert_eq!(
             self.version, other.version,
-            "attempting to merge projects with different versions"
+            "attempting to merge identifiers for projects with different versions"
         );
 
         for iri in &other.identifiers {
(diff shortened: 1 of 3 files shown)


daumantas-kavolis-sensmetry:
> This is communicated through a "resolved manifest" current.toml

`current.toml` is very non-descriptive or sounds quite irrelevant to an environment. `env.toml` or `projects.toml` would be far clearer.

> In particular this means paths are absolute

I highly suggest using paths relative to manifest/`sysand_env` for paths inside `sysand_env` and project folders as majority of the paths will have an identical prefix that can trivially be removed.

> What about IRIs for non-PURL usages? Since names will likely not be unique (e.g. 10 projects named "Requirements"), how will they be disambiguated to the user?

Why would anyone care about collisions in what is effectively a display name? If a user has multiple projects named the same, they have bigger issues than some ambiguous name.

daumantas-kavolis-sensmetry:
Maybe also add `workspace = [0, 1, ...]` top-level property that controls workspace/project roots for analysis?

core/src/env/local_directory/manifest.rs
```diff
@@ -0,0 +1,157 @@
...
+        utils::{FsIoError, wrapfs},
+    },
+};
+
+#[derive(Debug, Error)]
+pub enum ResolvedManifestError {
+    #[error(transparent)]
+    ResolutionError(#[from] ResolutionError<LocalReadError>),
+    #[error("too many dependencies, unable to convert to i64: {0}")]
+    TooManyDependencies(TryFromIntError),
```
daumantas-kavolis-sensmetry: I don't think this is achievable in practice as memory would run out first. If an integer conversion fails, it's far more likely because the manifest is corrupted. Additionally, negative values should also be an error.

```suggestion
    #[error("invalid dependency index, expected an unsigned integer: {0}")]
    InvalidDependencyIndex(TryFromIntError),
```

core/src/env/local_directory/metadata.rs
```diff
@@ -0,0 +1,363 @@
...
+        local_src::{LocalSrcError, LocalSrcProject},
+        utils::{
+            FsIoError, ToUnixPathBuf, deserialize_optional_unix_paths, deserialize_unix_path,
+            relativize_path, wrapfs,
+        },
+    },
+};
+
+pub const METADATA_PREFIX: &str = "# This file is automatically generated by Sysand and is not intended to be edited manually.\n\n";
+pub const CURRENT_METADATA_VERSION: &str = "0.1";
```
andrius-puksta-sensmetry: Should probably use an int here and for the lockfile. That can wait though.

core/src/env/local_directory/metadata.rs
```diff
@@ -0,0 +1,363 @@
...
+    /// The first identifier is to. be considered the canonical
+    /// identifier, and if the project is not `editable` this
+    /// is the IRI it is installed as. The rest are considered
+    /// as aliases. Can only be empty for `editable` projects.
+    #[serde(skip_serializing_if = "Vec::is_empty", default)]
+    pub identifiers: Vec<String>,
+    /// Usages of the project. Intended for tools needing to
+    /// track the interdependence of project in the environment.
+    #[serde(skip_serializing_if = "Vec::is_empty", default)]
+    pub usages: Vec<String>,
```
andrius-puksta-sensmetry: To make usage resolution faster and simpler for others, I agree with Daumantas that providing list indices would be better here. AFAIK `toml`/`toml_edit` support ordered serialization/deserialization.

Providing indices is much more annoying on our side though...

sysand/src/lib.rs
```diff
@@ -147,6 +147,7 @@ pub fn run_cli(args: cli::Args) -> Result<()> {
     let ctx = ProjectContext {
         current_workspace: discover_workspace(&cwd)?,
         current_project: discover_project(&cwd)?,
+        current_directory: cwd.clone(),
```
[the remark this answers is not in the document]
victor-linroth-sensmetry: I don't want to juggle the lifetimes required for this, so just transferred the ownership.

PR #205, opened 2026-03-12
reference: ai fix of regression in main

consideRatio:
Observed regression in main branch, mentioned in #206, is potentially fixed by these changes, generated by codex GPT 5.4. The text below is AI generated, and the changes are as well. I haven't looked into this myself. I have only verified that this change does unbreak `sysand add urn:kpar:sysmod` specifically.

---

The regression was in the lock path, not in add itself. The latest commit made lock generation require project.sources(), and embedded InMemoryProject instances for stdlib dependencies still hard-panicked there. I fixed that in memory.rs and populated concrete RemoteKpar sources for embedded stdlib projects in stdlib.rs.

Verification:

- cargo test -p sysand-core stdlib_projects_have_nominal_sources
- Exact repro, with network enabled for the index lookup:
  ```
  d=$(mktemp -d) && cd "$d" && cargo run --manifest-path /home/erik/dev/sensmetry/sysand/Cargo.toml init && cargo run --manifest-path /home/erik/dev/sensmetry/sysand/Cargo.toml add urn:kpar:sysmod
  ```

That flow now succeeds, creates sysand-lock.toml, and installs urn:kpar:sysmod 5.0.0-alpha.2 instead of panicking.

### DIFF
--- a/core/src/project/memory.rs
+++ b/core/src/project/memory.rs
@@ -157,6 +157,6 @@ impl ProjectRead for InMemoryProject {
     }
 
     fn sources(&self) -> Vec<Source> {
-        panic!("`InMemoryProject` cannot have any project sources")
+        self.nominal_sources.clone()
     }
 }
--- a/core/src/stdlib.rs
+++ b/core/src/stdlib.rs
@@ -1,5 +1,6 @@
 use std::collections::HashMap;
 
+use crate::lock::Source;
 use crate::project::memory::InMemoryProject;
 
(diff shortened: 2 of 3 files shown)


andrius-puksta-sensmetry (changes requested):
Ok, except duplicates have to be cleaned up.

andrius-puksta-sensmetry:
After thinking a bit, in my opinion it would be better to not even try including sources for `provided_iris`, which would avoid this problem.

Issue #210, opened 2026-03-13
Index rework overview


consideRatio:
<details><summary>Done / Preliminary done tasks</summary>

1. DONE
   **Add client support for publisher field**
   - Respect a `publisher` field in `.project.json`
   - Fixed by #212

2. PRELIMINARY DONE
   **Define index structure**
   - Fixed by https://gitlab.com/sensmetry/internal2/tech/syside/sysand/home/-/merge_requests/1
   - Fixed by #288

   The index structure has been defined like this, and it is different from the environment structure.
   ```
    # index_root
    # ├──index.json:
    # │    {
    # │      "projects": [
    # │        {
    # │          i = "pkg:sysand/publisher1/name1",
    # │          v = ["0.0.1", "0.0.2"]
    # │        }
    # │      ]
    # │    }
    # └──publisher1
    #     └──name1
    #         ├──0.0.1
    #         ├──0.0.2
    #         └──meta.json:
    #             {
    #                 "iri": "pkg:sysand/publisher1/name1",
    #                 "versions": ["0.0.1", "0.0.2"]
    #             }
   ```

3. PRELIMINARY DONE
   **Define index client->server upload/publish API**
   A preliminary strategy is implemented in the [sysand-index](https://gitlab.com/sensmetry/internal2/tech/syside/sysand/index-website), sending a POST request with multipart form-data, including a PURL and the .kpar data blob (with bundled README).

4. DONE
   **Add README.md bundling to build command**
   - Resolved by #86

5. DONE
   **Add client publish command**
   - #249

6. DONE
   **Change client's index reading behavior (breaking)**
   - Tracked by #286
   - Fixed by #288

7. DONE
   **Decisions about index server paths**
   Currently beta.sysand.org exposes things directly under its root path `/` (https://beta.sysand.org/entries.txt), but the new index currently doesn't (https://redacted.domain/index/index.json), it instead exposes it under `/index/`.

   A client reading the index should read from `https://redacted.domain/index/`, and a client publishing something should do it to `https://redacted.domain/` where `api/v1/upload` is appended.

   https://crates.io makes use of a config.json file where the client looks up where it should do various things, looking like https://github.com/rust-lang/crates.io-index/blob/master/config.json:

   ```json
   {
     "dl": "https://static.crates.io/crates",
     "api": "https://crates.io"
   }
   ```

[...]

PR #249, opened 2026-04-01
Add publish command

consideRatio:
- for https://github.com/sensmetry/sysand/issues/210

## Summary

This PR introduces a new `sysand publish` command to upload `.kpar` artifacts to a sysand package index.

## What’s Included

- Adds CLI command used as `sysand publish --index <URL> [PATH]` where `--index` is required
- Resolves `[PATH]` explicitly, or defaults to the output a build command would default to
- Makes use of a new API between client/server. A POST request with multipart form payload to `<index>/api/v1/upload`. It has two form fields.
  - `metadata` (application/json) with:
    - `normalized_publisher`
    - `normalized_name`
    - `version`
    - `license`
    - `kpar_sha256_digest`
  - `kpar` with the .kpar data blob (`application/zip`)
- Validates publish inputs before upload:
  - `publisher`/`name` must be normalizable based on strict rules
  - `version` must be valid SemVer 2.0
  - `license` must be valid SPDX expression
- Normalizes published ID fields (lowercase, spaces -> hyphens; dots in `name` are preserved)
- Validates Index URL:
  - only `http`/`https`
  - rejects query/fragment
  - rejects URLs that already include `/api/v1/upload`
- Makes use of publish-specific auth behavior:
  - only bearer-token credentials are used
  - credential matching is performed against the resolved upload URL
  - clear failures for missing/ambiguous bearer credentials
- Maps HTTP errors to user-friendly CLI errors (`401/403/404/409`)

## Docs

- Adds `docs/src/commands/publish.md`
- Adds `sysand publish` entry to docs summary navigation

## Tests

- Adds dedicated core publish tests (URL handling, validation)
- Adds extensive CLI integration tests for success paths and failure modes
- New tests added

## Dependency/Config Updates

- Enables `multipart` features for `reqwest` / `reqwest-middleware` where needed for upload support

### DIFF
--- a/core/src/commands/build.rs
+++ b/core/src/commands/build.rs
@@ -1,4 +1,4 @@
-use camino::Utf8Path;
+use camino::{Utf8Path, Utf8PathBuf};
 use thiserror::Error;
 
 use crate::{
@@ -207,6 +207,19 @@ impl<ProjectReadError: ErrorBound> From<IntoKparError<LocalSrcError>>
             IntoKparError::Serialize(msg, e) => Self::Serialize(msg, e),
         }
     }
+}
+
+pub fn default_kpar_path<Pr: ProjectRead>(
+    project: &Pr,
+    workspace: Option<&Workspace>,
+    project_path: &Utf8Path,
+) -> Result<Utf8PathBuf, KParBuildError<Pr::Error>> {
+    let mut path = workspace
+        .map(Workspace::root_path)
+        .unwrap_or(project_path)
+        .join("output");
+    path.push(default_kpar_file_name(project)?);
+    Ok(path)
 }
 
 pub fn default_kpar_file_name<Pr: ProjectRead>(
--- a/core/src/commands/publish.rs
+++ b/core/src/commands/publish.rs
@@ -193,6 +193,16 @@ fn prepare_publish_payload(kpar_path: &Utf8Path) -> Result<PublishPreparation, P
         .unwrap_or(kpar_path.as_str())
         .to_string();
 
+    let file_size = std::fs::metadata(kpar_path)
+        .map_err(|e| PublishError::KparRead(kpar_path.as_str().into(), e))?
+        .len();
+    if file_size > MAX_KPAR_PUBLISH_SIZE {
+        return Err(PublishError::KparTooLarge {
+            size: file_size,
+            limit: MAX_KPAR_PUBLISH_SIZE,
+        });
+    }
+
     let file_bytes = std::fs::read(kpar_path)
         .map_err(|e| PublishError::KparRead(kpar_path.as_str().into(), e))?;
     let sha256_digest = format!("{:x}", sha2::Sha256::digest(&file_bytes));
@@ -317,6 +327,11 @@ pub enum PublishError {
 
     #[error("publish endpoint not found: {0}")]
     NotFound(String),
+
+    #[error(
+        "kpar file is unexpectedly large ({size} bytes, limit is {limit} bytes); verify you are publishing the correct file"
+    )]
+    KparTooLarge { size: u64, limit: u64 },
 }
 
 #[derive(Debug)]
@@ -326,6 +341,8 @@ pub struct PublishResponse {
     pub is_new_project: bool,
 }
 
+/// Defensive upper bound on kpar file size (100 MiB) to catch unexpected uploads by mistake.
+const MAX_KPAR_PUBLISH_SIZE: u64 = 100 * 1024 * 1024;
 /// Maximum number of characters to include when summarizing an error response body.
 const MAX_ERROR_BODY_CHARS: usize = 1024;
 /// Path segments appended to the index URL to form the upload endpoint.
--- a/core/src/commands/publish.rs
+++ b/core/src/commands/publish.rs
@@ -16,8 +16,6 @@ use crate::{
 
 /// Defensive upper bound on kpar file size (100 MiB) to catch unexpected uploads by mistake.
 const MAX_KPAR_PUBLISH_SIZE: u64 = 100 * 1024 * 1024;
-/// Maximum number of characters to include when summarizing an error response body.
-const MAX_ERROR_BODY_CHARS: usize = 1024;
 /// Path segments appended to the index URL to form the upload endpoint.
 const UPLOAD_ENDPOINT_SEGMENTS: [&str; 3] = ["api", "v1", "upload"];
 
@@ -126,9 +124,7 @@ pub fn build_upload_url(index: &Url) -> Result<Url, PublishError> {
     }
 
     {
-        let mut segments = upload_url
-            .path_segments_mut()
-            .unwrap();
+        let mut segments = upload_url.path_segments_mut().unwrap();
         for segment in UPLOAD_ENDPOINT_SEGMENTS {
             segments.push(segment);
         }
@@ -321,10 +317,10 @@ fn map_publish_response(
             message: String::from_utf8_lossy(body_bytes).into_owned(),
             is_new_project: true,
         }),
-        401 | 403 => Err(PublishError::AuthError(summarize_error_body(body_bytes))),
-        409 => Err(PublishError::Conflict(summarize_error_body(body_bytes))),
-        400 => Err(PublishError::BadRequest(summarize_error_body(body_bytes))),
-        404 => Err(PublishError::NotFound(summarize_error_body(body_bytes))),
+        401 | 403 => Err(PublishError::AuthError(error_body_to_string(body_bytes))),
+        409 => Err(PublishError::Conflict(error_body_to_string(body_bytes))),
+        400 => Err(PublishError::BadRequest(error_body_to_string(body_bytes))),
+        404 => Err(PublishError::NotFound(error_body_to_string(body_bytes))),
         _ => {
             log::warn!(
                 "publish failed: request URL `{}`, final URL `{}`, status {}",
@@ -334,7 +330,7 @@ fn map_publish_response(
             );
             Err(PublishError::ServerError {
                 status,
-                body: summarize_error_body(body_bytes),
+                body: error_body_to_string(body_bytes),
             })
         }
     }
@@ -392,61 +388,18 @@ fn normalize_field(s: &str) -> String {
     s.to_ascii_lowercase().replace(' ', "-")
 }
 
-fn summarize_error_body(body_bytes: &[u8]) -> String {
+fn error_body_to_string(body_bytes: &[u8]) -> String {
     if body_bytes.is_empty() {
         return "empty response body".to_string();
     }
 
-    if let Ok(json) = serde_json::from_slice::<serde_json::Value>(body_bytes) {
-        let error = json.get("error").and_then(|v| v.as_str());
-        let detail = json.get("detail").and_then(|v| v.as_str());
-        let message = match (error, detail) {
-            (Some(error), Some(detail)) => format!("{error}: {detail}"),
-            (Some(error), None) => error.to_string(),
-            (None, Some(detail)) => detail.to_string(),
-            (None, None) => String::new(),
-        };
-        if !message.is_empty() {
-            return summarize_error_text(&message);
-        }
-    }
-
-    match std::str::from_utf8(body_bytes) {
-        Ok(text) => {
-            if text.chars().any(|c| c.is_control() && !c.is_whitespace()) {
-                return format!(
-                    "unexpected non-text error response ({} bytes)",
-                    body_bytes.len()
-                );
-            }
-            summarize_error_text(text)
-        }
-        Err(_) => format!(
-            "unexpected non-text error response ({} bytes)",
-            body_bytes.len()
-        ),
-    }
-}
-
-fn summarize_error_text(text: &str) -> String {
+    let text = String::from_utf8_lossy(body_bytes);
     let trimmed = text.trim();
     if trimmed.is_empty() {
-        return "empty response body".to_string();
+        "empty response body".to_string()
+    } else {
+        trimmed.to_string()
     }
-
-    let mut summarized = trimmed.to_string();
-    if summarized.len() > MAX_ERROR_BODY_CHARS {
-        let mut cutoff = MAX_ERROR_BODY_CHARS;
-        while !summarized.is_char_boundary(cutoff) {
-            cutoff -= 1;
-        }
-        summarized.truncate(cutoff);
-    }
-    if summarized.len() < trimmed.len() {
-        summarized.push_str(" ... [truncated]");
-    }
-
-    summarized
 }
 
 #[cfg(test)]
--- a/core/src/commands/publish.rs
+++ b/core/src/commands/publish.rs
@@ -59,7 +59,7 @@ pub fn do_publish<P: AsRef<Utf8Path>>(
 
         let form = reqwest::multipart::Form::new()
             .part("metadata", metadata_part)
-            .part("file", file_part);
+            .part("kpar", file_part);
 
         c.post(upload_url.clone()).multipart(form)
     };
@@ -104,7 +104,7 @@ pub fn build_upload_url(index: &Url) -> Result<Url, PublishError> {
         });
     }
 
-    let mut upload_url = index.clone();
+    let mut upload_url = index.to_owned();
     {
         let mut segments = upload_url
             .path_segments_mut()
@@ -128,7 +128,7 @@ pub fn build_upload_url(index: &Url) -> Result<Url, PublishError> {
     {
         let mut segments = upload_url
             .path_segments_mut()
-            .expect("http(s) URLs are hierarchical and must support mutable path segments");
+            .unwrap();
         for segment in UPLOAD_ENDPOINT_SEGMENTS {
             segments.push(segment);
         }
@@ -193,10 +193,10 @@ pub enum PublishError {
     #[error("invalid index URL `{url}` for publish endpoint: {reason}")]
     InvalidIndexUrl { url: Box<str>, reason: String },
 
-    #[error("HTTP request failed: {0}")]
+    #[error("HTTP request failed: {0:#?}")]
     Http(#[from] reqwest_middleware::Error),
 
-    #[error("failed to read server response body: {0}")]
+    #[error("failed to read server response body: {0:#?}")]
     ResponseBody(#[source] reqwest::Error),
 
     #[error("server error ({status}): {body}")]
@@ -215,7 +215,7 @@ pub enum PublishError {
     NotFound(String),
 
     #[error(
-        "kpar file is unexpectedly large ({size} bytes, limit is {limit} bytes); verify you are publishing the correct file"
+        "KPAR file is unexpectedly large ({size} bytes, limit is {limit} bytes); verify you are publishing the correct file"
     )]
     KparTooLarge { size: u64, limit: u64 },
 }
@@ -243,7 +243,7 @@ fn prepare_publish_payload(path: &Utf8Path) -> Result<PublishPreparation, Publis
 
     let info = info.ok_or(PublishError::MissingInfo)?;
     // Validate that metadata exists; contents are not used during upload.
-    let _meta = meta.ok_or(PublishError::MissingMeta)?;
+    _ = meta.ok_or(PublishError::MissingMeta)?;
 
     let publisher = info
         .publisher
--- a/core/src/commands/publish.rs
+++ b/core/src/commands/publish.rs
@@ -48,14 +48,14 @@ pub fn do_publish<P: AsRef<Utf8Path>>(
     let build_request = move |c: &reqwest_middleware::ClientWithMiddleware| {
         let metadata_part = reqwest::multipart::Part::text(metadata.clone())
             .mime_str("application/json")
-            .expect("hard-coded content type must be a valid MIME");
+            .unwrap();
         let kpar_part = reqwest::multipart::Part::stream(kpar_bytes.clone())
             // we declare an arbitrary filename to help server side libraries
             // make reasonable assumptions reading the POST request, such as not
             // trying to parse the binary data as UTF-8 or similar
             .file_name("project.kpar")
             .mime_str("application/zip")
-            .expect("hard-coded content type must be a valid MIME");
+            .unwrap();
 
         let form = reqwest::multipart::Form::new()
             .part("metadata", metadata_part)
--- a/core/src/commands/publish.rs
+++ b/core/src/commands/publish.rs
@@ -340,63 +340,6 @@ fn map_publish_response(
     }
 }
 
-fn summarize_error_text(text: &str) -> String {
-    let trimmed = text.trim();
-    if trimmed.is_empty() {
-        return "empty response body".to_string();
-    }
-
-    let mut summarized = trimmed.to_string();
-    if summarized.len() > MAX_ERROR_BODY_CHARS {
-        let mut cutoff = MAX_ERROR_BODY_CHARS;
-        while !summarized.is_char_boundary(cutoff) {
-            cutoff -= 1;
-        }
-        summarized.truncate(cutoff);
-    }
-    if summarized.len() < trimmed.len() {
-        summarized.push_str(" ... [truncated]");
-    }
-
-    summarized
-}
-
-fn summarize_error_body(body_bytes: &[u8]) -> String {
-    if body_bytes.is_empty() {
-        return "empty response body".to_string();
-    }
-
-    if let Ok(json) = serde_json::from_slice::<serde_json::Value>(body_bytes) {
-        let error = json.get("error").and_then(|v| v.as_str());
-        let detail = json.get("detail").and_then(|v| v.as_str());
-        let message = match (error, detail) {
-            (Some(error), Some(detail)) => format!("{error}: {detail}"),
-            (Some(error), None) => error.to_string(),
-            (None, Some(detail)) => detail.to_string(),
-            (None, None) => String::new(),
-        };
-        if !message.is_empty() {
-            return summarize_error_text(&message);
-        }
-    }
-
-    match std::str::from_utf8(body_bytes) {
-        Ok(text) => {
-            if text.chars().any(|c| c.is_control() && !c.is_whitespace()) {
-                return format!(
-                    "unexpected non-text error response ({} bytes)",
-                    body_bytes.len()
-                );
-            }
-            summarize_error_text(text)
-        }
-        Err(_) => format!(
-            "unexpected non-text error response ({} bytes)",
-            body_bytes.len()
-        ),
-    }
-}
-
 /// Validates a publisher or name field for modern project IDs.
 ///
 /// Rules: 3-50 ASCII alphanumeric characters, with single separators (space,
@@ -447,6 +390,63 @@ fn is_valid_name(s: &str) -> bool {
 
 fn normalize_field(s: &str) -> String {
     s.to_ascii_lowercase().replace(' ', "-")
+}
+
+fn summarize_error_body(body_bytes: &[u8]) -> String {
+    if body_bytes.is_empty() {
+        return "empty response body".to_string();
+    }
+
+    if let Ok(json) = serde_json::from_slice::<serde_json::Value>(body_bytes) {
+        let error = json.get("error").and_then(|v| v.as_str());
+        let detail = json.get("detail").and_then(|v| v.as_str());
+        let message = match (error, detail) {
+            (Some(error), Some(detail)) => format!("{error}: {detail}"),
+            (Some(error), None) => error.to_string(),
+            (None, Some(detail)) => detail.to_string(),
+            (None, None) => String::new(),
+        };
+        if !message.is_empty() {
+            return summarize_error_text(&message);
+        }
+    }
+
+    match std::str::from_utf8(body_bytes) {
+        Ok(text) => {
+            if text.chars().any(|c| c.is_control() && !c.is_whitespace()) {
+                return format!(
+                    "unexpected non-text error response ({} bytes)",
+                    body_bytes.len()
+                );
+            }
+            summarize_error_text(text)
+        }
+        Err(_) => format!(
+            "unexpected non-text error response ({} bytes)",
+            body_bytes.len()
+        ),
+    }
+}
+
+fn summarize_error_text(text: &str) -> String {
+    let trimmed = text.trim();
+    if trimmed.is_empty() {
+        return "empty response body".to_string();
+    }
+
+    let mut summarized = trimmed.to_string();
+    if summarized.len() > MAX_ERROR_BODY_CHARS {
+        let mut cutoff = MAX_ERROR_BODY_CHARS;
+        while !summarized.is_char_boundary(cutoff) {
+            cutoff -= 1;
+        }
+        summarized.truncate(cutoff);
+    }
+    if summarized.len() < trimmed.len() {
+        summarized.push_str(" ... [truncated]");
+    }
+
+    summarized
 }
 
 #[cfg(test)]
--- a/core/src/commands/publish.rs
+++ b/core/src/commands/publish.rs
@@ -277,8 +277,9 @@ fn prepare_publish_payload(path: &Utf8Path) -> Result<PublishPreparation, Publis
     let kpar_bytes =
         std::fs::read(path).map_err(|e| PublishError::KparRead(path.as_str().into(), e))?;
     let sha256_digest = format!("{:x}", sha2::Sha256::digest(&kpar_bytes));
+    // The server derives project identity from the uploaded .kpar content.
+    // Client metadata only carries integrity data for now.
     let metadata = serde_json::json!({
-        "purl": purl_versioned,
         "sha256_digest": sha256_digest,
     })
     .to_string();
--- a/core/src/commands/publish.rs
+++ b/core/src/commands/publish.rs
@@ -304,10 +304,10 @@ fn map_publish_response(
             message: String::from_utf8_lossy(body_bytes).into_owned(),
             is_new_project: true,
         }),
-        401 | 403 => Err(PublishError::AuthError(error_body_to_string(body_bytes))),
-        409 => Err(PublishError::Conflict(error_body_to_string(body_bytes))),
         400 => Err(PublishError::BadRequest(error_body_to_string(body_bytes))),
+        401 | 403 => Err(PublishError::AuthError(error_body_to_string(body_bytes))),
         404 => Err(PublishError::NotFound(error_body_to_string(body_bytes))),
+        409 => Err(PublishError::Conflict(error_body_to_string(body_bytes))),
         _ => {
             log::warn!(
                 "publish failed: request URL `{}`, final URL `{}`, status {}",
--- a/core/src/commands/publish.rs
+++ b/core/src/commands/publish.rs
@@ -19,26 +19,25 @@ use crate::{
 mod tests;
 
 pub fn do_publish<P: AsRef<Utf8Path>>(
-    kpar_path: P,
-    index_url: Url,
+    path: P,
+    index: Url,
     auth: ForceBearerAuth,
     client: reqwest_middleware::ClientWithMiddleware,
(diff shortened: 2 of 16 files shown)


andrius-puksta-sensmetry:
> Uploads multipart form (`purl` + `file`) to `/api/v1/upload`

Why not use something like [Cargo publish protocol](https://doc.rust-lang.org/cargo/reference/registry-web-api.html#publish)?

> - Reads and validates project metadata from the `.kpar`:
>   - `publisher` present and canonicalizable
>   - `name` canonicalizable
>   - `version` valid SemVer 2.0

What about license? Should require it to be an SPDX expression. (EDIT: Agree! RESOLVED)

sysand/src/cli.rs
```diff
@@ -181,6 +181,20 @@ pub enum Command {
...
+        /// for a KPAR in the output directory with the current project's
+        /// name and version
+        #[clap(verbatim_doc_comment)]
+        path: Option<Utf8PathBuf>,
+
+        /// URL of the package index to publish to. Defaults to the
+        /// configured default index URL, otherwise the first configured
+        /// index URL, or https://beta.sysand.org
+        #[arg(long, verbatim_doc_comment)]
+        index: Option<String>,
```
```suggestion
        index: Option<Url>,
```

andrius-puksta-sensmetry:
> > Uploads multipart form (`purl` + `file`) to `/api/v1/upload`
> 
> Why not use something like [Cargo publish protocol](https://doc.rust-lang.org/cargo/reference/registry-web-api.html#publish)?

Alaternatively, why not be even simpler: pass the PURL in some header (e.g. `X-Sysand-publish-PURL`) and the KPAR as the `application/zip` body?



EDIT by Erik: Discussed a bit further in https://github.com/sensmetry/sysand/pull/249#issuecomment-4214660577 and https://github.com/sensmetry/sysand/pull/249#issuecomment-4222919164, we now have `metadata` as JSON and `kpar` as application/zip, where metadata includes a checksum and the purl.

core/src/commands/publish.rs
```diff
@@ -0,0 +1,341 @@
...
+        }),
+        201 => Ok(PublishResponse {
+            status,
+            message: body,
+            is_new_project: true,
+        }),
+        401 | 403 => Err(PublishError::AuthError(body)),
+        409 => Err(PublishError::Conflict(body)),
+        400 | 404 => Err(PublishError::BadRequest(body)),
+        _ => Err(PublishError::ServerError(status, body)),
```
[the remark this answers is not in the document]
andrius-puksta-sensmetry: Nevermind, index URL is already mentioned above.

core/src/commands/publish.rs
```diff
@@ -0,0 +1,341 @@
...
+#[cfg(test)]
+mod tests {
+    use super::{
+        PublishError, build_upload_url, canonicalize_modern_project_id_component,
+        is_canonicalizable_name_field_value, is_canonicalizable_publisher_field_value,
+    };
+
+    #[test]
+    fn publisher_field_canonicalizability() {
+        assert!(is_canonicalizable_publisher_field_value("Acme Labs"));
```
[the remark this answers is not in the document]
andrius-puksta-sensmetry: Also a case for `- ` (space and punctuation), and one or more cases of non-ASCII.

sysand/src/commands/publish.rs
```diff
@@ -0,0 +1,176 @@
...
+    index: Option<String>,
+    ctx: &ProjectContext,
+    config: &Config,
+    auth_policy: Arc<StandardHTTPAuthentication>,
+    client: reqwest_middleware::ClientWithMiddleware,
+    runtime: Arc<tokio::runtime::Runtime>,
+) -> Result<()> {
+    let kpar_path = resolve_publish_kpar_path(path, ctx)?;
+    if !kpar_path.is_file() {
+        bail!("kpar file not found at `{kpar_path}`, run `sysand build` first");
```
andrius-puksta-sensmetry:
```suggestion
        bail!("KPAR file not found at `{kpar_path}`, run `sysand build` first");
```

sysand/src/commands/publish.rs
```diff
@@ -0,0 +1,176 @@
...
+    path: Option<Utf8PathBuf>,
+    index: Option<String>,
+    ctx: &ProjectContext,
+    config: &Config,
+    auth_policy: Arc<StandardHTTPAuthentication>,
+    client: reqwest_middleware::ClientWithMiddleware,
+    runtime: Arc<tokio::runtime::Runtime>,
+) -> Result<()> {
+    let kpar_path = resolve_publish_kpar_path(path, ctx)?;
+    if !kpar_path.is_file() {
```
andrius-puksta-sensmetry: `wrapfs::is_file`

sysand/src/commands/publish.rs
```diff
@@ -0,0 +1,176 @@
...
+    if let Some(index_url) = config
+        .indexes
+        .iter()
+        .find(|index| index.default.unwrap_or(false))
+        .map(|index| index.url.as_str())
+        .or_else(|| config.indexes.first().map(|index| index.url.as_str()))
+    {
+        Url::parse(index_url).map_err(|e| anyhow!("invalid index URL in configuration: {e}"))
+    } else {
+        Ok(Url::parse(DEFAULT_INDEX_URL).expect("default publish index URL must be valid"))
```
andrius-puksta-sensmetry:
```suggestion
        Ok(Url::parse(DEFAULT_INDEX_URL).unwrap())
```
We know it's valid.

sysand/src/commands/publish.rs
```diff
@@ -0,0 +1,176 @@
...
+    }
```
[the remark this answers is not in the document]
andrius-puksta-sensmetry: I also don't like this fallback logic. For now we can just require explicit `--index` to always be provided, and figure out the fallbacks later.

For reference, `uv` publish logic (which appears to be different from download logic) seems to be:
1. index provided explicitly via flag (it must appear in config, since it's referenced by name, not URL)
2. `pypi.org`

See https://docs.astral.sh/uv/guides/package/#publishing-your-package

consideRatio:
@andrius-puksta-sensmetry I've now updated this PR after rework and more self-review, and I consider it ready to be reviewed again.

---

> Why not use something like [Cargo publish protocol](https://doc.rust-lang.org/cargo/reference/registry-web-api.html#publish)?
>
> Alaternatively, why not be even simpler: pass the PURL in some header (e.g. `X-Sysand-publish-PURL`) and the KPAR as the `application/zip` body?

We could go for cargo's publish protocol, but there wasn't a clear why / why not for me here so I picked use of multipart form as it felt a bit more straightforward then using crates binary protocol. I think it is good if we can ensure that we can include misc metadata though, so just having a single application/zip POST form with headers seems too rigid. I for example added `sha256_digest` to the metadata now so that we can verify that the client submitted what was intended to be submitted for example.

Do you want me to switch to Cargos binary protocol to deliver the metadata and file (but with different metadata content)? It also does the trick.

core/src/commands/publish.rs
```diff
@@ -0,0 +1,454 @@
...
+    #[error("missing project metadata in kpar")]
+    MissingMeta,
+
+    #[error("missing publisher in project info (required for publishing)")]
+    MissingPublisher,
+
+    #[error(
+        "publisher field `{0}` is invalid for modern project IDs: must be 3-50 characters, use only ASCII letters and numbers, may include single spaces or hyphens between words, and must start and end with a letter or number"
+    )]
+    InvalidPublisher(Box<str>),
```
Jonas-Puksta-Sensmetry, line 160:
```suggestion
    InvalidPublisher(String),
```
Is there a reason not to just use `String` here?

core/src/commands/publish.rs
```diff
@@ -0,0 +1,454 @@
...
+use thiserror::Error;
+use url::Url;
+
+use crate::{
+    auth::{ForceBearerAuth, HTTPAuthentication},
+    project::{ProjectRead, local_kpar::LocalKParProject},
+};
+
+/// Defensive upper bound on kpar file size (100 MiB) to catch unexpected uploads by mistake.
+const MAX_KPAR_PUBLISH_SIZE: u64 = 100 * 1024 * 1024;
```
Jonas-Puksta-Sensmetry, line 19: Currently Sysand doesn't support jsons in KPARs (I think), but they will be supported. And the sysml model jsons are really big, at least with the current zip compression, and I wouldn't be surprised if even a simple project will be more than 100 MiB.

This should be fine for now though.

core/src/commands/publish.rs
```diff
@@ -0,0 +1,454 @@
...
+    let kpar_project = LocalKParProject::new_guess_root(path)
+        .map_err(|e| PublishError::KparOpen(path.as_str().into(), e.to_string()))?;
+
+    let (info, meta) = kpar_project
+        .get_project()
+        .map_err(|e| PublishError::KparOpen(path.as_str().into(), e.to_string()))?;
+
+    let info = info.ok_or(PublishError::MissingInfo)?;
+    // Validate that metadata exists; contents are not used during upload.
+    let _meta = meta.ok_or(PublishError::MissingMeta)?;
```
Jonas-Puksta-Sensmetry:
```suggestion
    _ = meta.ok_or(PublishError::MissingMeta)?;
```

core/src/commands/publish.rs
```diff
@@ -0,0 +1,454 @@
...
+        _ => {
+            log::warn!(
+                "publish failed: request URL `{}`, final URL `{}`, status {}",
+                upload_url_for_log,
+                response_url,
+                status
+            );
+            Err(PublishError::ServerError {
+                status,
+                body: summarize_error_body(body_bytes),
```
Jonas-Puksta-Sensmetry: Do we want to return these error details to the user in all cases?
The answer might be obvious, but I'm not familiar enough with with this to know for myself. Often the "500 internal server error" message is deliberately lacking in details to not accidentally expose sensitive information.

core/src/commands/publish.rs
```diff
@@ -0,0 +1,454 @@
...
+}
+
+/// Validates a publisher or name field for modern project IDs.
+///
+/// Rules: 3-50 ASCII alphanumeric characters, with single separators (space,
+/// hyphen, and optionally dot when `allow_dot` is true) allowed between words.
+/// Must start and end with an alphanumeric character.
+///
+/// Publish-only; if additional surfaces need this, extract to a shared module.
+fn is_valid_field(s: &str, allow_dot: bool) -> bool {
```
Jonas-Puksta-Sensmetry, line 340: It would be nice to return a more precise error in case it is invalid. It would be better UX for people to immediately know what exactly sysand doesn't like about the name/publisher

core/src/commands/publish.rs
```diff
@@ -0,0 +1,454 @@
...
+    if let Ok(json) = serde_json::from_slice::<serde_json::Value>(body_bytes) {
+        let error = json.get("error").and_then(|v| v.as_str());
+        let detail = json.get("detail").and_then(|v| v.as_str());
+        let message = match (error, detail) {
+            (Some(error), Some(detail)) => format!("{error}: {detail}"),
+            (Some(error), None) => error.to_string(),
+            (None, Some(detail)) => detail.to_string(),
+            (None, None) => String::new(),
+        };
+        if !message.is_empty() {
```
Jonas-Puksta-Sensmetry: What happens if message is empty? The body was a valid json, so I feel like something should be returned from here either way, and not fall through to the non-json body handling

core/src/commands/publish.rs
```diff
@@ -0,0 +1,454 @@
...
+            "unexpected non-text error response ({} bytes)",
+            body_bytes.len()
+        ),
+    }
+}
+
+fn summarize_error_text(text: &str) -> String {
+    let trimmed = text.trim();
+    if trimmed.is_empty() {
+        return "empty response body".to_string();
```
Jonas-Puksta-Sensmetry: Is "empty response body" an error we want to show to the user? Wouldn't it be better to just return a generic description based on the error code (e.g. "Not Found" for 404)?

core/src/commands/publish.rs
```diff
@@ -0,0 +1,454 @@
...
+
+use crate::{
+    auth::{ForceBearerAuth, HTTPAuthentication},
+    project::{ProjectRead, local_kpar::LocalKParProject},
+};
+
+/// Defensive upper bound on kpar file size (100 MiB) to catch unexpected uploads by mistake.
+const MAX_KPAR_PUBLISH_SIZE: u64 = 100 * 1024 * 1024;
+/// Maximum number of characters to include when summarizing an error response body.
+const MAX_ERROR_BODY_CHARS: usize = 1024;
```
Jonas-Puksta-Sensmetry: Are there any known cases when the errors would be really long?

andrius-puksta-sensmetry:
> Do you want me to switch to Cargos binary protocol to deliver the metadata and file (but with different metadata content)? It also does the trick.

Your choice, I don't have a strong opinion either way. `multipart/form-data` imposes a slight performance penalty (for generating the field separator string), but is widely supported (which is IMO much more important).

Either way I think it would be better to have separate fields for `publisher`, `name` and `version`. It's also unclear to me if these are required to match the actual KPAR metadata (IMO they should be the same). If yes, combining them into a PURL can be done server side, as publisher/name/version will have to be checked anyway.

core/src/commands/publish.rs
andrius-puksta-sensmetry: To reduce the error size. `Box<str>` is 16 bytes, `String` is 24. Since errors are returned from many places and tend to be somewhat large, they dictate the size of the return types.

core/src/commands/publish.rs
```diff
@@ -0,0 +1,454 @@
...
+        name,
+        version,
+        file_name,
+        file_bytes,
+        metadata,
+    } = prepare_publish_payload(path)?;
+    log::info!(
+        "{header}{:>12}{header:#} `{name}` {version} to {index}",
+        "Publishing",
+    );
```
andrius-puksta-sensmetry, line 41: Maybe mention `publisher` as well?

core/src/commands/publish.rs
```diff
@@ -0,0 +1,454 @@
...
+
+    // Stash the URL as a string for post-request logging; the `Url` itself
+    // is moved into the closure since `with_authentication` may call it
+    // multiple times and each `post` consumes the URL.
+    let upload_url_for_log = upload_url.to_string();
+
+    let build_request = move |c: &reqwest_middleware::ClientWithMiddleware| {
+        let file_part = reqwest::multipart::Part::stream(file_bytes.clone())
+            .file_name(file_name.clone())
```
andrius-puksta-sensmetry: Why not use `Part::file(<path>)`? Also, `file_name` serves no purpose; `name` will already identify the part.

core/src/commands/publish.rs
```diff
@@ -0,0 +1,454 @@
...
+            .mime_str("application/zip")
+            .expect("hard-coded content type must be a valid MIME");
+        let metadata_part = reqwest::multipart::Part::text(metadata.clone())
+            .mime_str("application/json")
+            .expect("hard-coded content type must be a valid MIME");
+
+        let form = reqwest::multipart::Form::new()
+            .part("metadata", metadata_part)
+            .part("file", file_part);
```
andrius-puksta-sensmetry:
```suggestion
            .part("kpar", file_part);
```
More explicit.

core/src/commands/publish.rs
```diff
@@ -0,0 +1,454 @@
...
+    }
+
+    if index.fragment().is_some() {
+        return Err(PublishError::InvalidIndexUrl {
+            url: index.as_str().into(),
+            reason: "URL must not include a fragment component".to_string(),
+        });
+    }
+
+    let mut upload_url = index.clone();
```
andrius-puksta-sensmetry:
```suggestion
    let mut upload_url = index.to_owned();
```

core/src/commands/publish.rs
```diff
@@ -0,0 +1,454 @@
...
+    {
+        let mut segments = upload_url
+            .path_segments_mut()
+            .expect("http(s) URLs are hierarchical and must support mutable path segments");
```
andrius-puksta-sensmetry: No need for error messages, simple `unwrap` + comment is fine, this is guaranteed to not fail for http(s)

core/src/commands/publish.rs
```diff
@@ -0,0 +1,454 @@
...
+        // Normalize both `https://host` and `https://host/`.
+        segments.pop_if_empty();
+    }
+
+    // After normalization, reject URLs that already end with the upload path.
+    let path_segments: Vec<_> = upload_url
+        .path_segments()
+        .expect("http(s) URLs are hierarchical and must support path segments")
```
andrius-puksta-sensmetry: ~~Just use `segments` from before.~~
Nevermind, can't do this.

core/src/commands/publish.rs
```diff
@@ -0,0 +1,454 @@
...
+        .collect();
+    if path_segments.ends_with(&UPLOAD_ENDPOINT_SEGMENTS) {
```
andrius-puksta-sensmetry: This is too verbose and inefficient. Use `.path()` and unify `UPLOAD_ENDPOINT_SEGMENTS` into a single `&str`.

core/src/commands/publish.rs
```diff
@@ -0,0 +1,454 @@
...
+        return Err(PublishError::InvalidIndexUrl {
+            url: index.as_str().into(),
+            reason: "URL must point to the index root; do not include `/api/v1/upload`".to_string(),
+        });
+    }
+
+    {
+        let mut segments = upload_url
+            .path_segments_mut()
+            .expect("http(s) URLs are hierarchical and must support mutable path segments");
```
andrius-puksta-sensmetry:
```suggestion
            .unwrap();
```
Since this is the same op as above, no need to explain.

core/src/commands/publish.rs
andrius-puksta-sensmetry: (to be fair, here specifically this has no effect on error size)

core/src/commands/publish.rs
```diff
@@ -0,0 +1,454 @@
...
+    InvalidLicense {
+        license: Box<str>,
+        source: spdx::error::ParseError,
+    },
+
+    #[error("invalid index URL `{url}` for publish endpoint: {reason}")]
+    InvalidIndexUrl { url: Box<str>, reason: String },
+
+    #[error("HTTP request failed: {0}")]
+    Http(#[from] reqwest_middleware::Error),
```
andrius-puksta-sensmetry:
```suggestion
    #[error("HTTP request failed: {0:#?}")]
    Http(#[from] reqwest_middleware::Error),
```
Due to the way `reqwest` implements `Error`, crucial details get lost for `reqwest_middleware::Error` unless debug print is used.

core/src/commands/publish.rs
```diff
@@ -0,0 +1,454 @@
...
+
+    #[error("failed to read server response body: {0}")]
+    ResponseBody(#[source] reqwest::Error),
```
andrius-puksta-sensmetry: Same here
```suggestion
    #[error("failed to read server response body: {0:#?}")]
    ResponseBody(#[source] reqwest::Error),
```

core/src/commands/publish.rs
```diff
@@ -0,0 +1,454 @@
...
+    Conflict(String),
+
+    #[error("bad request: {0}")]
+    BadRequest(String),
+
+    #[error("publish endpoint not found: {0}")]
+    NotFound(String),
+
+    #[error(
+        "kpar file is unexpectedly large ({size} bytes, limit is {limit} bytes); verify you are publishing the correct file"
```
andrius-puksta-sensmetry:
```suggestion
        "KPAR file is unexpectedly large ({size} bytes, limit is {limit} bytes); verify you are publishing the correct file"
```
`KPAR` is an abbreviation

core/src/commands/publish.rs
andrius-puksta-sensmetry: I see, `stream` is to clone cheaply.

core/src/commands/publish.rs
Jonas-Puksta-Sensmetry: Does it matter to us if the enum type is a few bytes bigger? It seems like a micro optimization to me.

core/src/config/mod.rs
```diff
@@ -101,6 +101,9 @@ impl Config {
 #[derive(Debug, Default, Clone, PartialEq, Serialize, Deserialize)]
 pub struct Index {
     pub name: Option<String>,
+    // url is declared a String, but would ideally be declared an Url. However,
+    // that would come with challenges as Url provides no Default impl, which
+    // makes it impossible to use #[derive(Default)] on this struct.
     pub url: String,
```
andrius-puksta-sensmetry: Semantically doesn't make sense to implement `Default` for this. It only seems to be useful in a few tests. For convenience can expose function `from_url` that populates only `url` field.

core/src/commands/publish.rs
andrius-puksta-sensmetry: It is a micro optimization. Doesn't hurt though.

core/src/commands/publish.rs
consideRatio: No not really, this was too defensive coding! Stripped this logic

core/src/commands/publish.rs
consideRatio: I'll leave this as it is now as I see that the current code base does a mix of both Box<str> and String for its errors.

core/src/commands/publish.rs
consideRatio: I'm not 100% on this, but I think it can make sense to surface the error if the server passes it, expecting the server to not pass back sensitive details in an error message.

core/src/commands/publish.rs
consideRatio: The error is declaring exactly what is valid and not, but doesn't detail what part was broken currently.

I agree it could be nice, but I think we should keep it like this for now to avoid embracing the complexity for an ideal UX. In my mind, the ideal UX is to be presented with the rule and know all parts wasn't followed.

core/src/commands/publish.rs
consideRatio: Simplified things here and ensured the server side to return `{"error": "some message"}`. Here is how it looks now:

```rust
fn error_body_to_string(body_bytes: &[u8]) -> String {
    let text = String::from_utf8_lossy(body_bytes);
    let trimmed = text.trim();

    if trimmed.is_empty() {
        return "no error details provided".to_string();
    }

    serde_json::from_str::<ErrorResponse>(trimmed)
        .map(|error| error.error)
        .unwrap_or_else(|_| trimmed.to_string())
}
```

core/src/commands/publish.rs
consideRatio: The server side errors and associated messages, which should be guaranteed by our index servers.

  - 400 — Bad request (invalid metadata, missing fields, bad archive, digest mismatch, PURL validation)
  - 401 — Authentication failure (missing/invalid/expired token)
  - 403 — Authorization failure (wrong token scope, not a member, insufficient role)
  - 404 — Namespace not found
  - 409 — Version already exists

Otherwise the fallback is "no error details provided" for now.

core/src/commands/publish.rs
consideRatio: I've simplified things to not provide or have the server consider details about filename, which is what declaring `.file_name(...)` does here.

core/src/config/mod.rs
consideRatio: I've removed the comment, and not attempted to change `url: string` to a `url: Url` here or implement a `from_url` helper. This PR isn't working against this Index struct as it initially where.

core/src/commands/publish.rs
```diff
@@ -50,6 +50,7 @@ pub fn do_publish<P: AsRef<Utf8Path>>(
             .mime_str("application/json")
             .expect("hard-coded content type must be a valid MIME");
         let kpar_part = reqwest::multipart::Part::stream(kpar_bytes.clone())
+            .file_name("project.kpar")
```
consideRatio, line 56: The index server won't consider this value, but it's nice that it is set at all because of the Python library used to handle the incoming POST request provides a better API if it knows its a file rather than an arbitrary binary blob.

core/src/commands/publish.rs
```diff
@@ -0,0 +1,399 @@
...
+    // is moved into the closure since `with_authentication` may call it
+    // multiple times and each `post` consumes the URL.
+    let upload_url_for_log = upload_url.to_string();
+
+    let build_request = move |c: &reqwest_middleware::ClientWithMiddleware| {
+        let metadata_part = reqwest::multipart::Part::text(metadata.clone())
+            .mime_str("application/json")
+            .expect("hard-coded content type must be a valid MIME");
+        let kpar_part = reqwest::multipart::Part::stream(kpar_bytes.clone())
+            .file_name("project.kpar")
```
andrius-puksta-sensmetry, line 56: Why set this?

consideRatio: Commented in https://github.com/sensmetry/sysand/pull/249#discussion_r3072786189 - let's discuss it there.

core/src/commands/publish.rs
```diff
@@ -0,0 +1,399 @@
...
+        if b.is_ascii_alphanumeric() {
+            continue;
+        }
+
+        let is_separator = b == b'-' || b == b' ' || (allow_dot && b == b'.');
+        if !is_separator {
+            return false;
+        }
+
+        if !bytes[i - 1].is_ascii_alphanumeric() || !bytes[i + 1].is_ascii_alphanumeric() {
```
andrius-puksta-sensmetry: Why isn't checking the previous byte enough?

sysand/src/commands/publish.rs
```diff
@@ -0,0 +1,106 @@
...
+            bail!(
+                "`sysand publish` without [PATH] is not supported from a workspace root; \
+                 run the command from a project directory or pass an explicit .kpar path"
+            );
+        }
+        return Err(CliError::MissingProjectCurrentDir.into());
+    };
+
+    let mut path = ctx
+        .current_workspace
```
andrius-puksta-sensmetry: ~~Since we explicitly error out if workspace is present, this is a noop.~~
EDIT: This is needed if run from a project that is within a workspace, as output dir is shared.

sysand/src/commands/publish.rs
```diff
@@ -0,0 +1,106 @@
...
+    ctx: &ProjectContext,
+) -> Result<Utf8PathBuf> {
+    if let Some(path) = path {
+        return Ok(path);
+    }
+
+    // Without an explicit path, publish must resolve one concrete project artifact.
+    // From workspace root this is ambiguous, so require `[PATH]` there.
+    let Some(current_project) = ctx.current_project.as_ref() else {
+        if ctx.current_workspace.is_some() {
```
andrius-puksta-sensmetry: This will error out if any workspace is present, regardless if it's in the current dir or in some ancestor. Workspace/project is discovered the same way, by walking from current dir to root looking for them.

sysand/tests/cli_publish.rs
```diff
@@ -0,0 +1,601 @@
...
+    for pattern in expected_stderr {
+        assertion = assertion.stderr(predicate::str::contains(*pattern));
+    }
+    publish_mock.assert();
+
+    Ok(())
+}
+
+#[test]
+fn test_publish_401_maps_to_auth_error() -> TestResult {
```
andrius-puksta-sensmetry: Remove `test_` prefixes, names are already very long and `#[test]` is explicit. We are currently inconsistent with this, some places use `test_`, others not, but I would like to eliminate it.

consideRatio: Ah okay. Just to verify my understanding: we should be systematic about using `#[test]`, and not lean on `test_` prefix?

consideRatio: Opened https://github.com/sensmetry/sysand/issues/280 to track that

docs/src/commands/publish.md
```diff
@@ -0,0 +1,84 @@
...
+`--index` is required for `sysand publish`.
+
+The package identifier used during publish is derived from project metadata.
+Before publishing, ensure `version`, `publisher`, `name`, and `license` follow
+these rules:
+
+- `version`: must be a valid Semantic Versioning 2.0 version.
+
+- `publisher`: 3-50 characters, ASCII letters and numbers only, with optional
+  single spaces or hyphens between words, and must start and end with a letter
```
andrius-puksta-sensmetry:
```suggestion
- `version`: must be a valid Semantic Versioning 2.0 version.
- `publisher`: 3-50 characters, ASCII letters and numbers only, with optional
  single spaces or hyphens between words, and must start and end with a letter
```

sysand/tests/cli_publish.rs
andrius-puksta-sensmetry: `#[test]` is required for test functions, it's a macro.

sysand/src/commands/publish.rs
andrius-puksta-sensmetry: This should check if cwd is the same as workspace root.

andrius-puksta-sensmetry:
Looked through everything. One question/suggestion still not addressed:

> I think it would be better to have separate fields for publisher, name and version. It's also unclear to me if these are required to match the actual KPAR metadata (IMO they should be the same). If yes, combining them into a PURL can be done server side, as publisher/name/version will have to be checked anyway.

consideRatio:
> Looked through everything. One question/suggestion still not addressed:
> 
> > I think it would be better to have separate fields for publisher, name and version. It's also unclear to me if these are required to match the actual KPAR metadata (IMO they should be the same). If yes, combining them into a PURL can be done server side, as publisher/name/version will have to be checked anyway.

They must match with the content. The idea of having the purl field was to do an authorization check before reading the .kpar archive, as it may be a bit more expensive operation, thinking about DoS protection etc.

With a purl field its clear we mean the normalized parts, do you want name and publisher to be normalized or not?

We could also just remove the purl field and go with it, because we won't process the request until the .kpar is received anyhow etc, so I think my DoS thinking wasn't very relevant.

Decision options:
1. Remove purl, just having checksum in metadata
2. Remove purl, but add publisher and name without normalization
3. Remove purl, but add normalized_publisher and normalized_name with normalization
4. Keep things as they are
5. Something else

---

I think I lean towards 1 if something, but more than anything I lean towards momentum so I'm fine with quickly adjusting to any decision

EDIT: I've pushed a commit implementing option 1, can switch to another decision still just wanted to avoid being idle.

sysand/src/commands/publish.rs
consideRatio: Refactored this function for readability, and updated the comment here as well to be clearer! See if you think b584504 makes sense now.

consideRatio:
I've [checked that this functionality works e2e](https://github.com/consideRatio/sysand-test-repo/actions/runs/24340573240/job/71088731641) against developed index server, so I consider it ready, but I can also adjust to a decision in [the comment above](https://github.com/sensmetry/sysand/pull/249#issuecomment-4236576196).

core/src/commands/publish.rs
```diff
@@ -0,0 +1,411 @@
...
+
+    // Stash the URL as a string for post-request logging; the `Url` itself
+    // is moved into the closure since `with_authentication` may call it
+    // multiple times and each `post` consumes the URL.
+    let upload_url_for_log = upload_url.to_string();
+
+    let build_request = move |c: &reqwest_middleware::ClientWithMiddleware| {
+        let metadata_part = reqwest::multipart::Part::text(metadata.clone())
+            .mime_str("application/json")
+            .expect("hard-coded content type must be a valid MIME");
```
andrius-puksta-sensmetry: IMO `unwrap()` should be used instead of `expect()` where we statically know that failure will not happen.

sysand/src/commands/publish.rs
```diff
@@ -0,0 +1,105 @@
...
+            );
+        }
+        (None, None) => return Err(CliError::MissingProjectCurrentDir.into()),
+    };
+
+    let output_file_name = default_kpar_file_name(current_project)?;
+    Ok(current_project
+        .project_path
+        .join("output")
+        .join(output_file_name))
```
andrius-puksta-sensmetry: If workspace is present, build artifacts (kpars) go to `$WORKSPACE_ROOT/output/`. Therefore the previous version of this that considered the workspace was correct.

andrius-puksta-sensmetry:
> we won't process the request until the .kpar is received anyhow etc, so I think my DoS thinking wasn't very relevant.

Is it feasible to process the request before fully receiving it (not now, just asking about whether this would be doable in the future if the need arises)? Even if not, it would still be nice to be able to quickly reject invalid requests (e.g. trying to publish to the wrong publisher/name, or especially publishing to a not-yet-existing/typoed package, which will happen for legitimate users) without unzipping the kpar. This doesn't prevent DoS by malicious users (they can simply include different names in the json than in the package, so we still must check the package), but most invalid requests come from legitimate users. Therefore I lean towards option 3 + version (to quickly detect duplicate publishes to the same version).

andrius-puksta-sensmetry:
> Even if not, it would still be nice to be able to quickly reject invalid requests

For this to be effective, we also have to require that metadata "field" comes before the file in the request.

sysand/src/commands/publish.rs
consideRatio: Aaah okay!

The various behavior changes based on .sysand.toml and .workspace.json files makes things quite a bit more complicated!

consideRatio:
@andrius-puksta-sensmetry comment above:

> For this to be effective, we also have to require that metadata "field" comes before the file in the request.

AI responses to misc questions:

> The multipart/form-data spec (RFC 7578) does not guarantee that servers, proxies, or client libraries preserve field order. Client libraries (requests, curl, etc.) may reorder fields, especially when mixing file and non-file parts.

> The WSGI server (gunicorn, uwsgi, etc.) reads the entire request body into memory (or a temp file for large uploads) before passing it to Django.

---

It is possible to have the server parse a stream-read a POST request and abort mid-read etc, but this is really sidestepping the conveniences of most webserver software we build upon. If we look for this level of robustness, I think we should go for the even more solid strategy is to not have the index server receive the file at all. You can implement a multi-step flow where you first have the server generate a pre-signed URL for the client where the client can upload directly to object storage, and then the client does it, and communicates back to the server referencing the initial upload request, whereby the server processes it etc.

However, we are now in the weeds of optimization beyond even what crates.io, npm, pypi, etc do, and it raises the bar in general on implementing an index.

andrius-puksta-sensmetry:
Yeah, it was just a question about feasibility. I would still prefer option 3 + version to be implemented to catch user errors without having to unzip `.project.json`.

core/src/commands/build.rs
```diff
@@ -209,6 +209,19 @@ impl<ProjectReadError: ErrorBound> From<IntoKparError<LocalSrcError>>
...
+    project_path: &Utf8Path,
+) -> Result<Utf8PathBuf, KParBuildError<Pr::Error>> {
+    let mut path = workspace
+        .map(Workspace::root_path)
+        .unwrap_or(project_path)
+        .join("output");
+    path.push(default_kpar_file_name(project)?);
+    Ok(path)
+}
+
```
consideRatio, line 224: I wanted to avoid breaking DRY principle, so put this the logic of concluding the path in the build command instead and referenced it in publish command.

core/src/commands/publish.rs
```diff
@@ -277,10 +277,12 @@ fn prepare_publish_payload(path: &Utf8Path) -> Result<PublishPreparation, Publis
...
     let sha256_digest = format!("{:x}", sha2::Sha256::digest(&kpar_bytes));
-    // The server derives project identity from the uploaded .kpar content.
-    // Client metadata only carries integrity data for now.
     let metadata = serde_json::json!({
-        "sha256_digest": sha256_digest,
+        "normalized_publisher": normalized_publisher,
+        "normalized_name": normalized_name,
+        "version": version,
+        "license": license,
+        "kpar_sha256_digest": sha256_digest,
```
consideRatio, line 285: I added license which is also checked early here, and renamed the digest to be more obvious what it was about.

andrius-puksta-sensmetry (approved):

PR #282, opened 2026-04-14
refactor: unit tests in dedicated files

consideRatio:
Migrated all inline `#[cfg(test)] mod tests { ... }` blocks to separate `*_tests.rs` files, referenced via `#[path]` like below:

```rust
#[cfg(test)]
#[path = "./<basename>_tests.rs"]
mod tests;
```

## Additional changes

- Moved `macro_tests` from `core/src/project/mod.rs` to integration test file `core/tests/project_derive.rs` (generic derive variants merged into existing tests).
- Renamed 3 inconsistent `mod test` to `mod tests` (`project/reference.rs`, `env/memory.rs`, `env/reqwest_http.rs`).

---

- fixes #240 systematically


bindings/java/src/exceptions_tests.rs
```diff
@@ -0,0 +1,92 @@
+// SPDX-FileCopyrightText: © 2025 Sysand contributors <opensource@sensmetry.com>
```
andrius-puksta-sensmetry: What is your view on copyright dates? Options:
1. bump them on every file we touch (e.g. every file in this PR) to 2026
2. have a range 2025-2026
3. don't include dates at all (not recommended by [REUSE](https://reuse.software/spec-3.3/#format-of-copyright-notices))

Probably the most correct option is 2, but it sure is annoying to maintain.

consideRatio: Extracted this to https://github.com/sensmetry/sysand/issues/283 to make discussion more visible

consideRatio: I pushed a commit adding SPDX headers, aligning with #284, without causing merge conflicts.

andrius-puksta-sensmetry (approved):

Issue #283, opened 2026-04-14
Regarding SPDX-FileCopyrightText headers in our files


consideRatio:
@andrius-puksta-sensmetry wrote in https://github.com/sensmetry/sysand/pull/282#discussion_r3079150134:

```rust
// SPDX-FileCopyrightText: © 2025 Sysand contributors <opensource@sensmetry.com>
```

>What is your view on copyright dates? Options:
>1. bump them on every file we touch (e.g. every file in this PR) to 2026
>2. have a range 2025-2026
>3. don't include dates at all (not recommended by [REUSE](https://reuse.software/spec-3.3/#format-of-copyright-notices))
>
>Probably the most correct option is 2, but it sure is annoying to maintain.

andrius-puksta-sensmetry:
Let's go with option 2 (`<year_start>-<year_end>` or just `<year>` for single) unless someone has a better suggestion.

consideRatio:
I lack background to what we look to accomplish with having the headers, and I'm not clear on the purpose of copyright generally.

Asked an AI about why we may do this currently:

> A top-level LICENSE file covers the repo as a whole, but per-file headers survive when files are copied, vendored, or extracted into other projects. They make license compliance unambiguous even outside the original repo context.

Is this the reason we do it? Is this a good enough reason to motivate the chore work?

If this is worth doing, then my take is that we should automate it. There is a script in the repo already related to it: `scripts/add_header.sh` but its undocumented and unreferenced 😱!

It was added here, and unchanged since, @vytautas-astrauskas-sensmetry can you chime in about it?

```
commit fae79762a6c6af971ea3fb39909e47177da9c078
Author: Vytautas Astrauskas <vytautas.astrauskas@sensmetry.com>
Date:   Wed Sep 17 17:40:19 2025 +0300

    Initial public commit
    
    Signed-off-by: Vytautas Astrauskas <vytautas.astrauskas@sensmetry.com>
```

consideRatio:
I opened https://github.com/sensmetry/sysand/pull/284 with automation and initial updates to systematically provide headers in the same format.

consideRatio:
I figure there isn't an action point remaining related to this, but it would be good to know if / why these are used for the future in new situations.

PR #284, opened 2026-04-14
prek: add reuse annotate hook, and fix SPDX headers

consideRatio:
This is the most suitable fix for #283 that I could come up with - we wouldn't have to think about the SPDX headers ourselves, because they are added automatically by prek when needed.

We get an initial copyright year set when a new file is created, and we don't update it over time etc. This seems to be fine based on AI ramblings, but I have no good idea about the impact of updating a copyright year etc.


andrius-puksta-sensmetry (approved):
Good enough for now

PR #297, opened 2026-04-22
feat!: use new local environment structure

### LINKED ISSUE #5: More transparent directory structure in local environments and index
The current structure used in both `sysand_env` and for remote HTTP indices is the following

```
/entries.txt
/<hash-of-IRI_1>/versions.txt
/<hash-of-IRI_1>/<version_1>.kar
/<hash-of-IRI_1>/<version_2>.kar
...
/<hash-of-IRI_2>/versions.txt
/<hash-of-IRI_2>/<version_1>.kar
...
```

Where `entries.txt` contains verbatim (as UTF-8 strings) entries `IRI_1, IRI_2, ...` and directories are hex-encoded SHA256-hashes of these IRIs considered as UTF-8 byte arrays. 

While straightforwardly safe, this representation has a number of usability issues:

1. the structure is difficult for humans to navigate in, for example, a graphical file browser or command line shell),
2. when opening or seeing a file path or URL it is difficult to tell what usage the corresponding file or archive corresponds to

We would like to have a mapping that alleviates the above issues *at least* for the most typical use cases:

1. projects identified by "name", using `urn:kpar:<name>` IRIs
2. projects identified by direct resolvable HTTP(S) URLs

andrius-puksta-sensmetry:
Transition to new local environment structure (proposed in [internal MR](https://gitlab.com/sensmetry/internal2/tech/syside/sysand/home/-/merge_requests/1)), which uses `env.toml` to describe the projects present in local environment, instead of the current `entries.txt`/`versions.txt`. Unpacked non-editable projects are now stored in directories inside `sysand_env/lib/`. Each project directory name is generated from its identifier and version.

`env.toml` was introduced in #175, which contains more details about its exact contents, but it so far mostly contained information already present in `entries.txt`/`versions.txt`.

This PR does not change what information (fields/projects) is stored in `env.toml`. It is now documented in `ARCHITECTURE.md`.

Old environment structure (with `entries.txt`/`versions.txt`) is still used by `LocalBrowserStorageEnvironment` (present only in WASM bindings). There are currently no plans to change that.

Closes #5.


bindings/py/Cargo.toml
```diff
@@ -14,7 +14,7 @@ homepage.workspace = true
 [features]
 default = ["extension-module", "abi3"]
 extension-module = ["pyo3/extension-module"]
-abi3 = ["pyo3/abi3", "pyo3/abi3-py38"]
+abi3 = ["pyo3/abi3", "pyo3/abi3-py310"]
```
consideRatio, line 17: Aha... I wondered about this before.

consideRatio:
Unassigned me from review, re-add me / ping me and I'll give it highest prio!

andrius-puksta-sensmetry:
Noticed a few important bugs.

andrius-puksta-sensmetry:
This is finally in good enough shape to be reviewed.

consideRatio:
@andrius-puksta-sensmetry could you describe the changes more explicitly on a high level in the PR description, ideally referencing an in-repo text detailing the new environment structure?

ARCHITECTURE.md doesn't do a good job yet, as the details of `env.toml` is omitted, and `current.toml` is mentioned without being found in the codebase - but is mentioned in [a linked GitLab MR](https://gitlab.com/sensmetry/internal2/tech/syside/sysand/home/-/merge_requests/1), which in turn links to #175 claimed to have added `current.toml`. I presume `current.toml` was renamed to `env.toml`.

---

Here is an example PR description that would be sufficient to allow me to understand and review better.

> This PR replaces a local environment structure (involving entries.txt/versions.txt) with a new structure (involving env.toml) now described in ARCHITECTURE.md. Installed projects are now stored as unpacked directories under sysand_env/lib/, with env.toml recording their identifiers, versions, paths, usages, and editable/workspace status.
>
>The old environment structure remains in the JS/WASM bindings still, to be migrated in the future.

core/src/env/local_directory/metadata.rs
```diff
@@ -160,44 +94,105 @@ impl EnvMetadata {
...
+        })
     }
 
-    pub fn add_project(&mut self, project: EnvProject) {
-        if let Some(found) = self.find_project(&project.identifiers, &project.version) {
-            self.projects[found].merge_identifiers(&project);
-        } else {
-            self.projects.push(project);
-        }
+    pub(super) fn find_project(&self, identifier: &str) -> impl Iterator<Item = &EnvProject> {
```
consideRatio: find_project -> return iterator is a readability concern, but its also relevant to align naming with many other find_project functions

comment can be dismissed without action, just flagging in case you have an action idea

core/src/env/local_directory/metadata.rs
```diff
@@ -160,44 +94,105 @@ impl EnvMetadata {
...
+            .filter(move |(_idx, p)| p.identifiers.iter().any(|iri| iri == identifier))
     }
 
-    pub fn remove_project<S: AsRef<str>, V: AsRef<str>>(&mut self, iri: S, version: Option<V>) {
-        if let Some(v) = version {
-            self.projects.retain(|p| {
-                p.version != v.as_ref() || !p.identifiers.iter().any(|i| i == iri.as_ref())
-            });
+    // /// Find all versions of all projects that have at least one of the given identifiers
+    // pub(super) fn find_project_any(
```
consideRatio: I'm in favor of removing this instead of leaving it in a commented out state, even if it could be relevant in the future.

core/src/env/local_directory/metadata.rs
```diff
@@ -265,7 +261,7 @@ pub struct EnvProject {
 }
 
 impl EnvProject {
-    pub fn to_toml(&self) -> Table {
+    pub(super) fn to_toml(&self) -> Table {
```
consideRatio, line 262: 👍 👍 👍 on reducing the scope when we can, nice! Long term maintenance with a large area of public functions etc can slow down future development a lot - or break things.

core/src/env/local_directory/utils.rs
```diff
@@ -3,56 +3,81 @@
...
-
-pub fn remove_dir_if_empty<P: AsRef<Utf8Path>>(path: P) -> Result<(), FsIoError> {
-    match fs::remove_dir(path.as_ref()) {
-        Err(err) if err.kind() == io::ErrorKind::DirectoryNotEmpty => Ok(()),
-        r => r.map_err(|e| FsIoError::RmDir(path.to_path_buf(), e)),
+use crate::project::utils::{FsIoError, ToPathBuf, wrapfs};
+
+/// Removes all files in the directory.
+/// All errors are ignored, but logged with `log::warn!()`.
+pub fn clean_dir<P: AsRef<Utf8Path>>(path: P) {
```
consideRatio, line 16: Great to see these doc comments - especially on a general purpose utility function!

core/src/env/local_directory/utils.rs
```diff
@@ -3,56 +3,81 @@
...
-
-    dirs.sort_by(|a, b| b.cmp(a));
-
-    for dir in dirs {
-        remove_dir_if_empty(&dir)?;
-    }
-
-    Ok(())
-}
+// pub fn remove_dir_if_empty<P: AsRef<Utf8Path>>(path: P) -> Result<(), FsIoError> {
```
consideRatio: Commented out code blocks does make it a bit easier to re-add it when needed, but it makes us carry some weight when unused as well.

My preference is to directly remove it, or move it out of sight - potentially leaving a function signature stub to know it can be re-added from an archive of sorts.

core/src/env/local_directory/utils.rs
```diff
@@ -149,13 +174,14 @@ pub(crate) fn move_fs_item<P: AsRef<Utf8Path>, Q: AsRef<Utf8Path>>(
...
 
-pub fn try_move_files(paths: &Vec<(&Utf8Path, &Utf8Path)>) -> Result<(), TryMoveError> {
+pub fn try_move_files(paths: &[(&Utf8Path, &Utf8Path)]) -> Result<(), TryMoveError> {
     let tempdir = camino_tempfile::tempdir()
         .map_err(|e| TryMoveError::RecoveredIO(FsIoError::CreateTempFile(e).into()))?;
 
     let mut last_err = None;
 
     // move source files out of the way
+    // TODO: why is this needed?
```
consideRatio: It seems that the function intent is to safeguard against creating an in-between state where you partially move files. Seems very reasonable to me to do.

Ideally, there is a doc comment for the function describing its intent.

core/src/context.rs
```diff
@@ -18,4 +21,7 @@ pub struct ProjectContext {
     /// Path to current directory
     #[cfg(feature = "filesystem")]
     pub current_directory: Utf8PathBuf,
+    /// `sysand_env`
```
consideRatio: A bit counter productive comment - indicated to me env was referencing a path or similar.

I'm not sure what would make sense, but probably something clarifying the Option part - when/why is it set and not set?

sysand/src/lib.rs
```diff
@@ -738,19 +736,27 @@ fn iri_or_path_to_iri(
     })
 }
 
-pub fn get_env(project_root: impl AsRef<Utf8Path>) -> Result<Option<LocalDirectoryEnvironment>> {
-    let environment_path = project_root.as_ref().join(DEFAULT_ENV_NAME);
-    let env = wrapfs::is_dir(&environment_path)?
-        .then_some(LocalDirectoryEnvironment { environment_path });
-    Ok(env)
+/// Create `sysand_env` at `root`
```
consideRatio:
```suggestion
```

We won't create a `sysand_env` folder by using this function, right? There is a `get_or_create_env` below though.

core/src/env/local_directory/mod.rs
```diff
@@ -1,78 +1,286 @@
...
+        let mut path_iter = IriVersionFilename::new(iri, version);
+        let mut candidate = path_iter.next_candidate();
+        while self.metadata.project_dir_exists(candidate) {
+            candidate = path_iter.next_candidate();
+        }
+        let mut path = String::from(path_iter);
+        path.insert_str(0, PROJECT_PATH_PREFIX);
+        path.into()
     }
 }
```
consideRatio, line 278: A bit worried about the complexity of having multiple paths for the same iri / version combination.

Is this complexity stemming from the TODO comment in iri_normalize not being addressed yet?

```
// TODO: use IRI canonicalization to deduplicate usages:
// - canonicalize each IRI before trying to resolve
// - canonicalize each usage before adding
// - use canonical versions in `sysand-lock.toml` and `env.toml` (for identifiers
//   and usages)
```

bindings/py/Cargo.toml
andrius-puksta-sensmetry: Oversight. Missed when upgrading minimum python version.

consideRatio:
For any handling of relative paths, its nice to also check that what it resolves to is under an expected directory. This is otherwise often a security issue in projects. Flagged by codex as below.

2. High: corrupt `env.toml` paths can make uninstall delete outside
   `sysand_env`.

   Location: `core/src/env/local_directory/metadata.rs:237`,
   `core/src/env/local_directory/mod.rs:492`,
   `core/src/env/local_directory/utils.rs:16`.

   Non-editable `EnvProject.path` is documented as relative to the env
   directory, but deserialization does not enforce that it stays under
   `sysand_env/lib` or even under `sysand_env`. Uninstall computes
   `self.root_dir.join(project.path.as_str())` and recursively cleans that
   directory. A crafted or corrupted `env.toml` with `path = "../victim"` can
   delete sibling files outside the env.

   Why it matters: `env.toml` is persisted local state, and corruption or manual
   edits should not let an uninstall remove arbitrary neighboring directories.

   Suggested fix direction: validate metadata paths on read and before delete.
   For non-editable projects, reject absolute paths, `..`, root/prefix
   components, and paths outside the owned project directory prefix. Prefer
   canonical containment checks before recursive deletion.

   Suggested verification: add a negative test with `path = "../victim"` and
   assert uninstall errors without touching the sibling directory.

core/src/env/local_directory/metadata.rs
andrius-puksta-sensmetry: Specifically here, it's force non-`LocalDirectoryEnvironment` to use methods exposed by `LocalDirectoryEnvironment`.

core/src/env/local_directory/utils.rs
andrius-puksta-sensmetry: In this case, I'd prefer to keep it, as it's quite likely it will be needed in the future, and the interface/logic is unlikely to change much, so moved all commented functions to end of file.

core/src/env/local_directory/utils.rs
andrius-puksta-sensmetry: Function intent is indeed to prevent partial moves, but I don't get why this specific loop is needed. The logic of the whole function is, given pairs of `(src, dest)`:

1. move every `src` file to a temp dir, revert on any error
2. move every `dest` file (if present) to a temp dir, revert on any error
3. move every file from temp src dir to `dest`
4. on error, revert everything.

I get why steps 2, 3 and 4 are needed, but what purpose does step 1 serve?

core/src/env/local_directory/mod.rs
andrius-puksta-sensmetry: > A bit worried about the complexity of having multiple paths for the same iri / version combination

Since the algorithm that turns IRI+version into a dir name is lossy, i.e. it will normalize `http://example.org/a` and `http://example.org/a--` into the same, we need a way to avoid duplicate names. Currently the logic is simply to insert a number after IRI into the generated dir name and increment it as long as duplicates exist.

> Is this complexity stemming from the TODO comment in iri_normalize not being addressed yet?

These are unrelated.

core/src/env/local_directory/utils.rs
consideRatio: I think the concrete case for relocating `src` files first is for situations when two different `(src, dest)` pairs overlap, so that a previous `dest` is a subsequent pairs `src` for example.

consideRatio (approved):
Nice! I think we should go for it whenever you are ready - no merge blockers on my end!

PR #298, opened 2026-04-22
feat: don't update metadata by default when building projects

andrius-puksta-sensmetry:
[Internal ref](https://gitlab.com/sensmetry/internal2/tech/syside/sysand/client/-/work_items/77)

- Don't update index symbols and file checksums by default when building a kpar.
- Add CLI option `--update-meta` that re-indexes and re-hashes all files mentioned in either `meta.index` or `meta.checksum`
- Increase stack size on Windows to prevent stack overflows (likely related to [clap issue](https://github.com/clap-rs/clap/issues/5134))

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -9,7 +9,7 @@
 # crt-static reference: See https://rust-lang.github.io/rfcs/1721-crt-static.html
 # dumpbin verification: https://github.com/sensmetry/sysand/pull/91#issuecomment-4081523020
 #
-# Increase the stack size, since Windows default 1 MiB causes a stack
+# Increase the stack size to 8MiB, since Windows default 1 MiB causes a stack
 # overflow on almost any CLI operation
 [target.'cfg(windows)']
 rustflags = ["-Ctarget-feature=+crt-static", "-Clink-args=/STACK:8388608"]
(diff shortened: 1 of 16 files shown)


.cargo/config.toml
```diff
@@ -9,5 +9,7 @@
 # crt-static reference: See https://rust-lang.github.io/rfcs/1721-crt-static.html
 # dumpbin verification: https://github.com/sensmetry/sysand/pull/91#issuecomment-4081523020
 #
+# Increase the stack size, since Windows default 1 MiB causes a stack
+# overflow on almost any CLI operation
```
consideRatio:
```suggestion
# Increase the stack size to 8MiB, since Windows default 1 MiB causes a stack
# overflow on almost any CLI operation
```

sysand/tests/cli_add_remove.rs
```diff
@@ -519,6 +519,7 @@ fn add_and_remove_as_local_kpar() -> Result<(), Box<dyn std::error::Error>> {
         ["init", "--version", "1.2.3", "--name", "add_and_remove"],
         None,
     )?;
+    dbg!("hello");
```
consideRatio: hello?! :D

consideRatio (approved):
Nice! Looks right and in alignment with the https://gitlab.com/sensmetry/internal2/tech/syside/sysand/client/-/work_items/77 that mentions specifically that _a warning_ should be emitted.

Just emitting a warning though delays something that maybe always should have been an error though - should this really be just a warning? Opting out of erroring with a flag could also be an option, like `--no-semver` for example.

I don't think its critical to resolve this now, so going for "approve". If you think its worth iterating on, maybe you can open an issue to track what you think makes sense to do, unless its done pre-merge of course.

PR #314, opened 2026-04-30
ci: release procedure rework

### LINKED ISSUE #83: Packaging release assets
1. Currently, release assets are uncompressed, see https://github.com/sensmetry/sysand/releases/tag/0.0.4 for example. I would expect them to be compressed to reduce download sizes, `.tar.xz` (preferably) or `.tar.gz` on Linux/macOS, and `.zip` on Windows (because it only recently gained support for `.tar.xz` some users may not have up-to-date systems).
2. Executables should not have abi tags, only the downloaded archives. Switching between platforms I would not want to run `sysand-linux-x64` on Linux and `sysand-macos-arm64`, I would expect that `sysand` would just work on both platforms.

consideRatio:
This PR makes us get a single moving `nightly` tag and single nightly GitHub Release, which together with [a GitLab MR](https://gitlab.com/sensmetry/internal2/tech/syside/sysand/sysand-signing/-/merge_requests/2) should help us promote it to stable releases. This was requested by @andrius-puksta-sensmetry but I don't think there is an issue about it.

Also, with this, we start providing archives of the binaries instead of the binaries directly, and that fixes #83.

Our release artifacts will be archives with an associated executable marked binary:
- `sysand-linux-arm64.tar.xz` - `sysand`
- `sysand-linux-x86_64.tar.xz` - `sysand`
- `sysand-macos-arm64.tar.xz` - `sysand`
- `sysand-macos-x86_64.tar.xz` - `sysand`
- `sysand-windows-arm64.zip` - `sysand.exe`
- `sysand-windows-x86_64.zip` - `sysand.exe`

---

I've tested this at [consideratio/sysand](https://github.com/consideRatio/sysand/actions) together with [the GitLab MR](https://gitlab.com/sensmetry/internal2/tech/syside/sysand/sysand-signing/-/merge_requests/2), but not the `sign` command, only `download`, `package`, `publish`.

### DIFF
--- a/.github/workflows/java.yml
+++ b/.github/workflows/java.yml
@@ -3,7 +3,7 @@ name: Java (Bindings)
 on:
   push:
     branches: [main]
-    tags: ["*"]
+    tags: ["v*"]
   pull_request:
     branches: [main]
   workflow_dispatch:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -197,25 +197,19 @@ jobs:
         run: |
           NIGHTLY_TAG=nightly
 
-          if ! gh release view "$NIGHTLY_TAG" >/dev/null 2>&1; then
-            echo "GitHub release '$NIGHTLY_TAG' does not exist or cannot be accessed."
-            echo "Create it first, then rerun this workflow."
-            exit 1
+          if gh release view "$NIGHTLY_TAG" >/dev/null 2>&1; then
+            gh release delete "$NIGHTLY_TAG" --yes
           fi
-
-          for asset in $(gh release view "$NIGHTLY_TAG" --json assets --jq '.assets[].name'); do
-            gh release delete-asset "$NIGHTLY_TAG" "$asset" --yes
-          done
-
-          gh release upload "$NIGHTLY_TAG" artifacts/sysand-*
 
           git tag --force "$NIGHTLY_TAG" "$GITHUB_SHA"
           git push --force origin "refs/tags/$NIGHTLY_TAG"
 
-          gh release edit "$NIGHTLY_TAG" \
+          gh release create "$NIGHTLY_TAG" artifacts/sysand-* \
               --title "Nightly Release" \
               --prerelease \
-              --latest=false
+              --latest=false \
+              --verify-tag
+
         env:
           GH_TOKEN: ${{ github.token }}
           GH_REPO: ${{ github.repository }}
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -3,7 +3,7 @@ name: Rust (Library & CLI)
 on:
   push:
     branches: [main]
-    tags: ["*"]
+    tags: ["v*"]
   pull_request:
     branches: [main]
   workflow_dispatch:
@@ -172,7 +172,7 @@ jobs:
   nightly-release:
     name: Nightly Release
     needs: [test, build]
-    if: github.ref == 'refs/heads/main' && !startsWith(github.ref, 'refs/tags/')
+    if: github.ref == 'refs/heads/main'
     runs-on: ubuntu-24.04
     permissions:
       id-token: write # for actions/attest
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -120,13 +120,25 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os:
-          - ubuntu-24.04
-          - ubuntu-24.04-arm
-          - windows-latest
-          - windows-11-arm
-          - macos-15-intel # x86
-          - macos-latest # arm64
+        include:
+          - os: ubuntu-24.04
+            asset: sysand-linux-x86_64
+            ext: ""
+          - os: ubuntu-24.04-arm
+            asset: sysand-linux-arm64
+            ext: ""
+          - os: windows-latest
+            asset: sysand-windows-x86_64.exe
+            ext: .exe
+          - os: windows-11-arm
+            asset: sysand-windows-arm64.exe
+            ext: .exe
+          - os: macos-15-intel
+            asset: sysand-macos-x86_64
+            ext: ""
+          - os: macos-latest
+            asset: sysand-macos-arm64
+            ext: ""
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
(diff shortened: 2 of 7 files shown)


.github/workflows/rust.yml
```diff
@@ -180,65 +193,28 @@ jobs:
...
-              --title "Nightly Release $NEW_TAG" \
-              --prerelease \
-              --target "$GITHUB_SHA" \
-              artifacts/*
-        env:
-          GH_TOKEN: ${{ github.token }}
-          GH_REPO: ${{ github.repository }}
+          NIGHTLY_TAG=nightly
+          shopt -s nullglob
+          assets=(artifacts/sysand-linux-* artifacts/sysand-macos-* artifacts/sysand-windows-*)
```
andrius-puksta-sensmetry: IMO `shopt -s nullglob` here is harmful. If we somehow generate artifacts whose names don't match the glob, at least `gh release upload` below should fail and bring attention to the problem.

.github/workflows/rust.yml
```diff
@@ -180,65 +193,28 @@ jobs:
...
+          gh release view "$NIGHTLY_TAG" >/dev/null
 
-    steps:
-      - uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
-        with:
-          path: artifacts
-          merge-multiple: true
+          # Remove stale assets so the floating nightly release mirrors this run.
+          while read -r asset; do
+              gh release delete-asset "$NIGHTLY_TAG" "$asset" --yes
```
andrius-puksta-sensmetry: I don't understand how this works. `read` reads stdin, but nothing is piped to it, so it should block forever or (maybe) timeout.

.github/workflows/rust.yml
```diff
@@ -180,65 +193,28 @@ jobs:
...
-    name: Release
-    needs: [test, build]
-    if: startsWith(github.ref, 'refs/tags/')
-    runs-on: ubuntu-24.04
-    permissions:
-      id-token: write # for actions/attest
-      attestations: write # for actions/attest
-      artifact-metadata: write # for actions/attest
-      contents: write # for "gh release create"
+          gh release view "$NIGHTLY_TAG" >/dev/null
```
andrius-puksta-sensmetry: This will fail if the nightly release is deleted. Missing nightly should be tolerated.

andrius-puksta-sensmetry:
Wouldn't it be simpler to delete (potentially) existing nightly and create a new one, instead of:

- check that it exists
- delete each asset
- upload new assets
- update release name/description

.github/workflows/rust.yml
consideRatio: I had it like that but then iterated towards not including logic to handle a missing nightly release, as it reduced notable amonts of almost never used code, and instead manually created it for this repo.

Re-add it?

.github/workflows/rust.yml
```diff
@@ -180,65 +193,28 @@ jobs:
...
-          gh release create "$NEW_TAG" \
-              --title "Nightly Release $NEW_TAG" \
-              --prerelease \
-              --target "$GITHUB_SHA" \
-              artifacts/*
-        env:
-          GH_TOKEN: ${{ github.token }}
-          GH_REPO: ${{ github.repository }}
+          NIGHTLY_TAG=nightly
+          shopt -s nullglob
```
consideRatio:
```suggestion
```

consideRatio:
> Wouldn't it be simpler to delete (potentially) existing nightly and create a new one, instead of:
> 
> * check that it exists
> * delete each asset
> * upload new assets
> * update release name/description

Its a good idea, but its also not trivial to adjust to it. I started doing some work relevant doing anyhow for this in 26a52ff.

.github/workflows/rust.yml
consideRatio: I strengthened up this situation by code that errors clearly for now

.github/workflows/rust.yml
consideRatio: Its messy, I reduced the complexity of this code to be more readable and compromised on a detail that won't affect us (we now must not upload release artifacts with spaces in them).

.github/workflows/java.yml
```diff
@@ -3,7 +3,7 @@ name: Java (Bindings)
 on:
   push:
     branches: [main]
-    tags: ["*"]
+    tags: ["v*"]
```
andrius-puksta-sensmetry, line 6: This is not necessary, as releases created by an action and using default GH actions token are special - they don't trigger CI. This is also how our current method works - new nightly releases are published all the time, but they don't trigger CI reruns.

docs/src/getting_started/installation.md
```diff
@@ -213,9 +220,12 @@ You should see an output similar to: `sysand X.Y.Z`
...
 
-Latest development version of Sysand can be downloaded from
-[GitHub releases][gh_rel_all] by choosing the latest release by date
-(which is usually labelled starting with "Nightly Release").
+Latest development version of Sysand can be downloaded from the
+[`nightly` GitHub release][gh_rel_nightly].
+
+> [!warning]
+> Nightly assets are raw signing inputs for release automation. They are not
+> packaged like the official release downloads above.
```
andrius-puksta-sensmetry:
```suggestion
> [!warning]
> Nightly assets are not signed and so may be prevented from running by
> antivirus software.
```

.github/workflows/java.yml
consideRatio: Ah yes true, but this makes things a bit more readable still and sensible, right? We don't for example want to go for a pypi release on a tag not following the `v*` pattern, if someone pushed it manually.

.github/workflows/rust.yml
```diff
@@ -161,7 +172,7 @@ jobs:
   nightly-release:
     name: Nightly Release
     needs: [test, build]
-    if: github.ref == 'refs/heads/main' && !startsWith(github.ref, 'refs/tags/')
+    if: github.ref == 'refs/heads/main'
```
andrius-puksta-sensmetry, line 175: This means that nightly release will be made twice per each regular release: once after the last PR is merged and another when the release tag is created (as that tag is created using PAT and so causes CI re-run).

consideRatio: Hmmm, I don't understand why it would run when a tag is created, before or after this change.

`github.ref == 'refs/heads/main' && !startsWith(github.ref, 'refs/tags/')` was weird, because whenever the first part is true, the second part must be true, because we know github.ref doesn't start with refs/tags - it starts with refs/heads.

So this change was a no-change, removing a always-true second part.

RELEASE.md
```diff
@@ -0,0 +1,33 @@
+# Releasing Sysand
+
+The release procedure is coordinated with the
+[sysand-signing GitLab project][sysand-signing]. That project is the source of
+truth for signing, packaging, and publishing GitHub Release assets.
+
+At a high level, a Sysand version release starts in this repository:
+
+1. Prepare a pull request that updates the version number and any generated
+   lockfiles that must change with it.
```
andrius-puksta-sensmetry:
```suggestion
1. Prepare a pull request that updates the version number and any generated
   lockfiles that must change with it ([example](https://github.com/sensmetry/sysand/pull/310)).
```

andrius-puksta-sensmetry (approved):
With suggestion above applied, this is good to go. We should add all the generated artifacts to releases, but that is tracked by #317.

PR #336, opened 2026-05-13
feat!: rename `sysand_env` to `.sysand`

andrius-puksta-sensmetry:
Via `sed -i 's/sysand_env/.sysand/' <all files>`.

The only remaining `sysand_env` occurences are in the changelog.


consideRatio (approved):
We could add a notice about detecting an outdated sysand_env, but its OK to proceed without also!

PR #339, opened 2026-05-18
refactor: centralize sha256 calculations, update sha2 to 0.11 and gix to 0.83

andrius-puksta-sensmetry:
- upgrade `sha2` to 0.11
- upgrade `gix` to 0.83
- centralize SHA256 hash calculations and serializing to hex strings


Cargo.lock
andrius-puksta-sensmetry, line 1: This has such a large positive diff largely due to `gix` being split up into more and more crates, and different versions of `sha2` being pulled in.

consideRatio (approved):
Nice!!

PR #340, opened 2026-05-20
release: prepare 0.1.0-alpha.1

andrius-puksta-sensmetry:


consideRatio (approved):

consideRatio:
Whatever is meant for use externally in any degree would make sense to call an alpha / beta, but its not a big deal so let's go!

PR #341, opened 2026-05-20
docs: refine release instructions

consideRatio:
Signed-off-by: Erik Sundell <erik.sundell+2025@sensmetry.com>

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -118,6 +118,35 @@ based on the following rules:
 
 ### Bump version entries
 
-1. Update the `version` entry in `bindings/js/package.json`
-2. Update the `workspace.package.version` entry in `Cargo.toml`
-3. Run `cargo generate-lockfile`
+We look to bump our own declared versions, and do a last minute security audit
+of dependencies for known vulnerabilities and fix them if possible.
+
+1. Update `workspace.package.version` in `Cargo.toml`, and the lockfile by
+   doing:
+
+   ```sh
+   # updates lockfile minimalistically
+   cargo check
+
+   # do a last minute security audit
+   cargo audit
+
+   # and bump manually to address known vulnerabilities
+   cargo update -p vulnerable-crate-1 [--precise 1.2.3]
+   ```
+
+2. Update `version` in `bindings/js/package.json` manually, and the lockfile by
+   doing:
+
+   ```sh
+   cd bindings/js/
+
+   # updates lockfile minimalistically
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -118,6 +118,35 @@ based on the following rules:
...
-3. Run `cargo generate-lockfile`
+We look to bump our own declared versions, and do a last minute security audit
+of dependencies for known vulnerabilities and fix them if possible.
+
+1. Update `workspace.package.version` in `Cargo.toml`, and the lockfile by
+   doing:
+
+   ```sh
+   # updates lockfile minimalistically
+   cargo check
```
consideRatio, line 129: generate-lockfile re-updates everything to the latest version, but cargo check does the right amount of things I think.

So `cargo check` and `npm install` over `cargo generate-lockfile` and `npm update`.

RELEASE.md
```diff
@@ -118,6 +118,35 @@ based on the following rules:
...
+
+   # updates lockfile minimalistically
+   npm install --package-lock-only --ignore-scripts
+
+   # do a last minute security audit
+   npm audit
+
+   # and bump minimalistically to address known vulnerabilities
+   npm audit fix
+   ```
```
andrius-puksta-sensmetry, line 152: `uv lock` is also needed to update `uv.lock`.

RELEASE.md
andrius-puksta-sensmetry: Yup, just noticed this doing the current release...

RELEASE.md
consideRatio: I not sure, but I don't think `uv lock` would accomplish anything in our release process:

- We don't manually update pyproject.toml to have a new version - its set dynamically by the build tooling maturin, so `uv lock` won't change anything in the uv.lock file I think. `uv lock --upgrade` would, but thats not suitable to do during a release procedure I think.
- uv.lock represents our local build environment, compared to Cargo.toml / Cargo.lock and package.json / package-lock.json that is more like `pyproject.toml` and determines versions of software installed by end users. We don't have any dependencies directly under `pyproject.toml` except the `dev` dependencies which is captured by `uv lock` into the `uv.lock` file.

andrius-puksta-sensmetry: You're right, I somehow managed to forget that uv.lock doesn't change right after making a release PR myself...

andrius-puksta-sensmetry (approved):

PR #345, opened 2026-05-21
feat!: make standard library identifiers not use `urn:kpar:`

Jonas-Puksta-Sensmetry:


docs/src/config/dependencies.md
```diff
@@ -27,16 +27,16 @@ project.
...
 [[project]]
 identifiers = [
-    "urn:kpar:my-project",
+    "pkg:sysand/my-publisher/my-local-src-project",
 ]
 sources = [
-    { src_path = "path/to/project" },
+    { src_path = "pkg:sysand/my-publisher/local-project" },
 ]
 ```
```
consideRatio, line 41: This seems off, on line 39 i expected a local path still

andrius-puksta-sensmetry (approved):

PR #349, opened 2026-05-24
feat: add Java bindings for writing .project.json and .meta.json

vytautas-astrauskas-sensmetry:


consideRatio (approved):
Since this is Java only, I didnt actually look at code, but approve as to signal i think we should move fast and accept breaking things now/later when it comes to our bindings.

Ping me again and i'll prioritize review of this in detail over other work!

andrius-puksta-sensmetry (approved):

PR #352, opened 2026-05-25
feat!: make all KPAR sources have kpar_digest and src sources have checksum

### LINKED ISSUE #325: Remove project_digest from an index's `versions.json`
Update: we decided to remote `project_digest` field from index.

Discussion extracted from https://github.com/sensmetry/sysand/issues/308:

@andrius-puksta-sensmetry wrote in https://github.com/sensmetry/sysand/issues/308#issuecomment-4394871500:
> A separate option would be to remove `project_digest` from `versions.json` and incur two additional roundtrips to get `.project.json`/`.meta.json`. This would avoid having the contract altogether, and clients can calculate their checksums however they see fit.

@consideRatio wrote:
> But `project_digest` aka. `checksum` has relevance out of scope of the index interactions, right? For example, if you have a lockfile, and then you have some remote .kpar or other kind of project source you download on `sysand sync` using the lockfile.

@consideRatio wrote:

>So `lock` determines `project_digest` by inspecting the `.json` files, which it already does since we expose them in full.
>
>Image

@consideRatio wrote:

>While `lock` can still compute the digest, but we would no longer have a way to verify the downloaded json files were as intended.
>
>If you delegate downloading of files to mirrors etc, you often want to verify that they serve what they should. In the current sysand index server, access of versions.json is from the main webserver, while access of .meta.json and .project.json, and the .kpar, is from object storage.
>
>I trust this object storage as much as I trust the main webserver in this case, but we would by removing project_digest from versions.json lock in a requirement on trusting the backend serving the .meta.json and .project.json files. Maybe that is okay, but its the drawback I think of.

### LINKED ISSUE #326: Remove checksum in Project struct and introduce project_digest and kpar_digest fields in Source variants
From notes discussed during a video chat with Andrius, Jonas, Victor, and I.

```
3. make lockfile checksum field optional one way or another?
    YES, we now provide Source specific digests
    - Editable, RemoteGit --- no project_digest
    - LocalKpar, RemoteKpar, IndexKpar --- kpar_digest
    - LocalSrc, RemoteSrc --- project_digest
```

To be decided by the implementer, but I was thinking Source fields name and valued like:

```
[..._]project_digest: "sysand:<lowercase-hex>"
[..._]kpar_digest: "sha256:<lowercase-hex>"
```

andrius-puksta-sensmetry:
## Breaking changes

- Removed project-level `checksum` from lockfile project entries.
- Moved integrity metadata into individual lockfile sources:
  - `*_src` sources now all carry `checksum`
  - `kpar_path` / `*_kpar` sources now all carry `kpar_size` and `kpar_digest`
- Config source overrides now use a separate `OverrideSource` model, since config records user-provided source locations while lockfiles record resolved sources with integrity metadata.
- Bumped lockfile version, older lockfiles are incompatible.

## Added

  - Added `ProjectChecksum`, distinguishing canonical source-project checksums from KPAR archive checksums. It is tied rather closely to `env.toml` structure, so other `ReadEnvironment` implementers currrently stub out the methods.
  - Added environment checksum verification APIs so sync can check whether an installed project matches the checksum kind and value required by a lockfile source.
  - Added `src_cksum` / `kpar_cksum` metadata in local environment project entries to remember what kind of source the project was installed from. This means that currently we have no way to verify correctness of projects with `kpar_cksum`, as the actual KPAR is not saved
  - README.md and licenses are now installed in the env when installing projects

  ## Changed

  - `sync` now decides whether a project is already installed by checking each lockfile source checksum against the environment, rather than comparing a single project-level checksum.
  - `sync` installs projects using the checksum kind associated with the selected source.
  - KPAR sources now verify the archive digest and size before use when expected metadata is available.
  - Provided-project handling in `sync` now matches by project version instead of project checksum.

## Fixed

- crash that happens when `env.toml` is written and later another project is being installed

Closes #325.
Closes #326.


consideRatio:
Dumping a codex review for consideration, but haven't worked through things myself - I'm trying to finalize doing lots of things in parallel and figure this needs my full attention to do manually.

<details><summary>Codex prompt</summary>

```
› You are on a feature branch, where we'll squash commits etc when merging.

  The PR title is "feat!: make all KPAR sources have kpar_digest and src sources have checksum
  ", and its description is pasted below. Can you based on the description task a few subagents to review with specific scope, relevant based on the PR title /
  description?

  ## Breaking changes

  - Removed project-level `checksum` from lockfile project entries.
  - Moved integrity metadata into individual lockfile sources:
    - `*_src` sources now all carry `checksum`
    - `kpar_path` / `*_kpar` sources now all carry `kpar_size` and `kpar_digest`
  - Config source overrides now use a separate `OverrideSource` model, since config records user-provided source locations while lockfiles record resolved
  sources with integrity metadata.
  - Bumped lockfile version, older lockfiles are incompatible.

  ## Added

    - Added `ProjectChecksum`, distinguishing canonical source-project checksums from KPAR archive checksums. It is tied rather closely to `env.toml`
  structure, so other `ReadEnvironment` implementers currrently stub out the methods.
    - Added environment checksum verification APIs so sync can check whether an installed project matches the checksum kind and value required by a lockfile
  source.
    - Added `src_cksum` / `kpar_cksum` metadata in local environment project entries to remember what kind of source the project was installed from. This means
  that currently we have no way to verify correctness of projects with `kpar_cksum`, as the actual KPAR is not saved
    - README.md and licenses are now installed in the env when installing projects

    ## Changed

    - `sync` now decides whether a project is already installed by checking each lockfile source checksum against the environment, rather than comparing a
  single project-level checksum.
    - `sync` installs projects using the checksum kind associated with the selected source.
    - KPAR sources now verify the archive digest and size before use when expected metadata is available.
    - Provided-project handling in `sync` now matches by project version instead of project checksum.

  ## Fixed

  - crash that happens when `env.toml` is written and later another project is being installed

  Closes #325.
  Closes #326.
```

</details>

---

I tasked four subagents and closed them after collecting results. Consolidated review scopes and headline findings:

  High Priority

  - core/src/project/index_entry.rs:137 / core/src/index/model.rs:56: index sidecar .project.json / .meta.json metadata can be used without verification against
    the verified KPAR. This overlaps with stale docs still requiring project_digest in docs/src/index-protocol.md:236 and verification at :397. Decide whether
    to restore project_digest validation or make KPAR-derived metadata authoritative.
  - core/src/commands/sync.rs:372 / core/src/env/utils.rs:57: remote_src checksum verification and install can read different bytes, then record the trusted
    src_cksum for unverified installed content.
  - core/src/commands/lock.rs:291 / core/src/lock.rs:410: lock extension deduplicates via Project::hash_val(), but the hash ignores source identity and per-
    source integrity metadata, so changed source digests can be silently kept stale.

  Panics / Correctness

  - core/src/env/index.rs:635: has_version_verified_async is implemented as panic!().
  - core/src/commands/sync.rs:153: provided-project version matching unwraps nested version() results and can panic on malformed/incomplete provided metadata.
  - core/src/env/memory.rs:257 / :362: in-memory env drops the supplied ProjectChecksum and recomputes a source-project checksum, so KPAR-installed projects
    won’t verify as KPAR checksum matches.

  Format / Diagnostics

  - core/src/lock.rs:284 / :308: uppercase digest validation is inconsistent across checksum, local_kpar, remote_kpar, and index_kpar.
  - core/src/lock.rs:388 / :448: Project.name appears to have become required without being documented as part of the lockfile break.
  - core/src/commands/sync.rs:19 / :395: checksum mismatch errors drop expected/actual checksum kind and value.
  - core/src/lock.rs:183 / core/src/lock_tests.rs:726: stale “canonical project-digest” terminology now conflates checksum and kpar_digest.

andrius-puksta-sensmetry:
> core/src/commands/sync.rs:372 / core/src/env/utils.rs:57: remote_src checksum verification and install can read different bytes, then record the trusted src_cksum for unverified installed content.

This is by design, as remote src project files are read on-demand. Out of scope here, but definitely will have to address in the future.

> core/src/env/index.rs:635: has_version_verified_async is implemented as panic!().

Intentional

> core/src/commands/sync.rs:153: provided-project version matching unwraps nested version() results and can panic on malformed/incomplete provided metadata.

Intentional

> core/src/env/memory.rs:257 / :362: in-memory env drops the supplied ProjectChecksum and recomputes a source-project checksum, so KPAR-installed projects
won’t verify as KPAR checksum matches.

Current limitation of `LocalBrowserStorageEnvironment`.

> core/src/commands/lock.rs:291 / core/src/lock.rs:410: lock extension deduplicates via Project::hash_val(), but the hash ignores source identity and per-
source integrity metadata, so changed source digests can be silently kept stale.

Partly fixed, partly intentional. It's not critical that deduplication works perfectly here, it just has to catch the obvious cases, which will otherwise cause the lock operation to fail due to duplicate exported symbols. Changed to deduplicate based on canonical (i.e. first) identifier. Currently the only way (AFAIK) for any duplication is with circular dependencies, which can only be encountered for the clone command due to the way the lock is constructed (root project is added first, so in case a project has circular deps on some other project is cloned, it will both be added to lock initially and from the dependency resolver).

consideRatio (approved):
While this is a massive change and I perceive I can't review this technically very well, i perceive its a scoped PR with intent that is fine.

I'd like us to lean on @andrius-puksta-sensmetry judgement if this is OK to merge at this point, and iterate and refine over time if needed.

consideRatio:
This PR will now close #326 which mentions introducing project_digest in favor of checksum, but this PR doesn't rename checksum to project_digest.

An explicit decision on doing that or not is relevant, and I figure its fine if you make the call alone @andrius-puksta-sensmetry, but I figure such rename shouldn't burden this PR and is better a separate PR if so.

PR #354, opened 2026-05-27
feat: support sysand add/remove PURL shorthand

consideRatio:
Users can identify indexed packages as `publisher/name`, but
`sysand add` previously required the full
`pkg:sysand/publisher/name` IRI. The CLI also parsed the positional
add locator as an IRI before core add logic could apply add-specific
behavior.

Expand valid Sysand PURL shorthand in the add path so normalized
`publisher/name` input is stored as the canonical PURL. Reject shorthand
that would require normalization first. Leave other add resources to
existing usage validation. Keep the shorthand scoped to add by leaving
general project usage validation and remove parsing unchanged. Update
CLI parsing, docs, and Python binding coverage for the new add behavior.

### DIFF
--- a/core/src/commands/add.rs
+++ b/core/src/commands/add.rs
@@ -7,6 +7,7 @@ use crate::{
         InterchangeProjectUsageG, InterchangeProjectUsageRaw, InterchangeProjectValidationError,
     },
     project::ProjectMut,
+    purl::{PKG_SYSAND_PREFIX, SysandPurlError, parse_sysand_purl},
 };
 
 const SP: char = ' ';
@@ -21,12 +22,42 @@ pub enum AddError<ProjectError> {
     MissingInfo(&'static str),
 }
 
+pub fn expand_sysand_purl_shorthand(
+    resource: &str,
+) -> Result<String, InterchangeProjectValidationError> {
+    let mut parts = resource.split('/');
+    let publisher = parts.next();
+    let name = parts.next();
+    let has_exactly_two_segments = publisher.is_some() && name.is_some() && parts.next().is_none();
+
+    // IRI always starts with `scheme:`, so differentiate from it by absence of `:`
+    if !resource.contains(':') && has_exactly_two_segments {
+        let purl = format!("{PKG_SYSAND_PREFIX}{resource}");
+        match parse_sysand_purl(&purl) {
+            Ok(Some(_)) => Ok(purl),
+            Err(source @ SysandPurlError::NotNormalized { .. }) => {
+                Err(InterchangeProjectValidationError::MalformedSysandPurl {
+                    iri: resource.to_owned(),
+                    source,
+                })
+            }
+            Ok(None) | Err(_) => Ok(resource.to_owned()),
+        }
+    } else {
+        Ok(resource.to_owned())
+    }
+}
+
 /// Ok(true) => usage added to project info
 /// Ok(false) => usage already present in project info
 pub fn do_add<P: ProjectMut>(
     project: &mut P,
     usage_raw: &InterchangeProjectUsageRaw,
 ) -> Result<bool, AddError<P::Error>> {
+    let usage_raw = InterchangeProjectUsageRaw {
+        resource: expand_sysand_purl_shorthand(&usage_raw.resource)?,
+        version_constraint: usage_raw.version_constraint.clone(),
+    };
     let usage: InterchangeProjectUsageG<String, String> = usage_raw.validate()?.into();
 
     let adding = "Adding";
@@ -109,3 +140,7 @@ pub fn do_add<P: ProjectMut>(
         ))
     }
 }
+
+#[cfg(test)]
+#[path = "./add_tests.rs"]
+mod tests;
--- a/core/src/commands/add.rs
+++ b/core/src/commands/add.rs
@@ -22,9 +22,11 @@ pub enum AddError<ProjectError> {
     MissingInfo(&'static str),
 }
 
-pub fn expand_sysand_purl_shorthand(
-    resource: &str,
-) -> Result<String, InterchangeProjectValidationError> {
+/// If `resource` is of shape `publisher/name`, and both satisfy Sysand PURL
+/// rules, return `Ok(Some(pkg:sysand/publisher/name))`. Otherwise, if it's of shape
+/// `string1/string2` and does not contain `:`, return error. If none of these,
+/// return `Ok(None)`, which indicates that it's likely something else
+pub fn expand_sysand_purl_shorthand(resource: &str) -> Result<Option<String>, SysandPurlError> {
     let mut parts = resource.split('/');
     let publisher = parts.next();
     let name = parts.next();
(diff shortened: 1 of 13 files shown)


core/src/commands/add.rs
```diff
@@ -21,12 +22,41 @@ pub enum AddError<ProjectError> {
...
 
+pub fn expand_sysand_purl_shorthand(
+    resource: &str,
+) -> Result<String, InterchangeProjectValidationError> {
+    let mut parts = resource.split('/');
+    let publisher = parts.next();
+    let name = parts.next();
+    let has_exactly_two_segments = publisher.is_some() && name.is_some() && parts.next().is_none();
+
+    if !resource.contains(':') && has_exactly_two_segments {
```
andrius-puksta-sensmetry, line 36:
```suggestion
    // IRI always starts with `scheme:`, so differentiate from it by absence of `:`
    if !resource.contains(':') && has_exactly_two_segments {
```

core/src/commands/add_tests.rs
andrius-puksta-sensmetry, line 1: Add a test that IRIs continue to be treated as IRIs.

sysand/src/commands/add.rs
```diff
@@ -47,6 +47,9 @@ pub fn command_add<Policy: HTTPAuthentication>(
     runtime: Arc<tokio::runtime::Runtime>,
     auth_policy: Arc<Policy>,
 ) -> Result<()> {
+    let iri = expand_sysand_purl_shorthand(&resource)?;
+    let iri = Iri::parse(iri)
+        .map_err(|(err, val)| anyhow::anyhow!("failed to parse `{val}` as IRI: {err}"))?;
```
andrius-puksta-sensmetry:
```suggestion
    let iri = Iri::parse(iri).expect("BUG: sysand PURL is invalid IRI");
```

andrius-puksta-sensmetry:
I think `sysand remove` should also have this to preserve symmetry.

core/src/commands/add.rs
```diff
@@ -21,12 +22,42 @@ pub enum AddError<ProjectError> {
...
 /// Ok(true) => usage added to project info
 /// Ok(false) => usage already present in project info
 pub fn do_add<P: ProjectMut>(
     project: &mut P,
     usage_raw: &InterchangeProjectUsageRaw,
 ) -> Result<bool, AddError<P::Error>> {
+    let usage_raw = InterchangeProjectUsageRaw {
+        resource: expand_sysand_purl_shorthand(&usage_raw.resource)?,
+        version_constraint: usage_raw.version_constraint.clone(),
+    };
```
andrius-puksta-sensmetry: I think this doesn't belong in the main path. Maybe split out into a separate wrapper function `fn do_add_shorthand`, as this is only needed for bindings.

sysand/src/commands/add.rs
```diff
@@ -47,6 +47,8 @@ pub fn command_add<Policy: HTTPAuthentication>(
     runtime: Arc<tokio::runtime::Runtime>,
     auth_policy: Arc<Policy>,
 ) -> Result<()> {
```
andrius-puksta-sensmetry, line 49: It would be better to still take IRI here. Coercing paths/shorthands to IRIs is better done by the CLI to have unified error messages, since this shorthand only has easily-checked syntactic requirements.

sysand/src/lib.rs
```diff
@@ -750,20 +750,18 @@ pub fn run_cli(args: cli::Args) -> Result<()> {
     }
 }
 
-fn iri_or_path_to_iri(
-    iri: Option<Iri<String>>,
+fn usage_locator_to_resource(
```
andrius-puksta-sensmetry: Make this return IRI as before by handling PURL shorthand here.

sysand/src/commands/add.rs
consideRatio: Hmmm, so you want bindings to use the core functionality by being explicit with `pkg:sysand/...` while the CLI can get the shorthand functionality?

My initial inclination is that this is behavior on the core level rather than CLI level, and having different CLI / bindings behavior would be unexpected.

core/src/commands/add_tests.rs
```diff
@@ -0,0 +1,79 @@
...
+    );
+}
+
+#[test]
+fn purl_shorthand_expansion_keeps_iri_resource() {
+    assert_eq!(
+        expand_sysand_purl_shorthand("https://example.com/acme-labs/my.project").unwrap(),
+        "https://example.com/acme-labs/my.project"
+    );
+}
```
andrius-puksta-sensmetry, line 41: Add tests that `do_add` and `do_remove` (preferably their wrappers) continue to work for IRIs.

sysand/src/cli.rs
```diff
@@ -1718,10 +1720,3 @@ impl ValueEnum for MetamodelVersion {
         })
     }
 }
-
-fn parse_iri_suggest_path(s: &str) -> Result<Iri<String>, String> {
-    use crate::style::USAGE;
-    Iri::parse(s.to_owned()).map_err(|(err, _val)| {
-        format!("{err}\n{USAGE}hint:{USAGE:#} if you wanted to use a path, use `--path` instead")
-    })
-}
```
andrius-puksta-sensmetry, line 1727: I would very much like to not lose this. Moving shorthand expansion here would allow to keep this.

consideRatio:
Hmmm this became more work than I have time to allocate to make deadlines. Do you have time to finalize/re-work it?

Decision required about core change vs CLI change:
a) We adjust add/remove in core, influencing all surfaces
b) We adjust add/remove in CLI only

My take is that this change isn't CLI specific, so it should be implemented in the core layer, but what we do matters less to me than we move forward.

If we would reprompt AI to implement this from scratch, I'd describe it as we want add/remove functionality in the core layer updated to:
- we ensure normalized publisher/name strings expands to pkg:sysand/publisher/name
- we error clearly on _normalizable but not normalized_ publisher/name
- we otherwise don't influence any other behavior (CLI help for "if you meant to use a path" etc is retained)

sysand/src/lib.rs
```diff
@@ -750,20 +751,20 @@ pub fn run_cli(args: cli::Args) -> Result<()> {
...
-    iri: Option<Iri<String>>,
+fn usage_locator_to_iri(
+    iri: Option<String>,
     path: Option<Utf8PathBuf>,
 ) -> Result<Iri<String>, anyhow::Error> {
     Ok(if let Some(iri) = iri {
-        iri
+        let iri = expand_sysand_purl_shorthand(&iri)?;
+        Iri::parse(iri).expect("BUG: usage locator is invalid IRI")
     } else {
```
andrius-puksta-sensmetry, line 759: `expand_sysand_purl_shorthand` is now called three times during the same add operation: when parsing CLI args, in `command_add` and in `do_add`.

I'll fix this.

sysand/src/commands/add.rs
andrius-puksta-sensmetry: I would rather keep the core functions clean. Shorthands and other non-functional convenience based on guessing (in core) should be done in dedicated functions (like `do_add_guess` I just added) or not at all. Otherwise it's difficult to determine the intent of the caller and make good errors/suggestions to the user.

andrius-puksta-sensmetry:
Done everything to my liking. Now core `do_add`/`do_remove` don't do guessing, there are wrapper functions `do_add_guess` and `do_remove_guess` for this. CLI verifies/parses the PURL once.

andrius-puksta-sensmetry (approved):

PR #362, opened 2026-05-29
feat: add Dockerfile and automation to publish on/for releases

### LINKED ISSUE #21: Create and publish a container image
It would be useful to have a canonical container image. The main use case would be in piplines that generate and publish KPar archives from a project repository.

consideRatio:
The intent of the dockerfile is to be used in automation where you
perhaps want to run sysand publish and acquire a trusted publishing
token from an index using curl.

Signed-off-by: Erik Sundell <erik.sundell@sensmetry.com>

### DIFF
--- a/.github/workflows/publish-image.yml
+++ b/.github/workflows/publish-image.yml
@@ -1,0 +1,95 @@
+name: Publish Image
+
+on:
+  release:
+    types: [created]
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: Git tag associated with the release to publish
+        required: true
+        type: string
+      use_default_branch_dockerfile:
+        description: Use the Dockerfile from the default branch instead of the release tag
+        required: false
+        type: boolean
+        default: false
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  publish-image:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          ref: ${{ inputs.use_default_branch_dockerfile && github.event.repository.default_branch || inputs.tag || github.event.release.tag_name }}
+
+      - name: Declare image tag
+        id: image
+        env:
+          TAG: ${{ inputs.tag || github.event.release.tag_name }}
+        run: echo "tag=${TAG#v}" >> "$GITHUB_OUTPUT"
+
+      - name: Wait for release assets to be uploaded
+        env:
+          GH_TOKEN: ${{ github.token }}
+          TAG: ${{ inputs.tag || github.event.release.tag_name }}
+        run: |
+          for attempt in {1..10}; do
+            assets="$(gh release view "$TAG" --repo "$GITHUB_REPOSITORY" --json assets --jq '.assets[].name')"
+
+            has_amd64_asset="$(grep -Fxc "sysand-linux-x86_64.tar.xz" <<< "$assets")"
+            has_arm64_asset="$(grep -Fxc "sysand-linux-arm64.tar.xz" <<< "$assets")"
+            if [[ "$has_amd64_asset" == 1 && "$has_arm64_asset" == 1 ]]; then
+              break
+            fi
+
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-image.yml
```diff
@@ -0,0 +1,95 @@
...
+
+            if [[ "$attempt" == 10 ]]; then
+              echo "Release $TAG does not include sysand-linux-x86_64.tar.xz and sysand-linux-arm64.tar.xz assets." >&2
+              exit 1
+            fi
+
+            echo "Release $TAG does not include all image assets yet. Sleeping before attempt $((attempt + 1))..."
+            sleep 6
+          done
+
```
andrius-puksta-sensmetry, line 58: Is there no better way? This looks hacky.

Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+# temporary trusted publishing tokens, relevant when using sysand publish in a
+# CI context
+RUN apt-get update \
+ && apt-get install -y --no-install-recommends \
+        ca-certificates \
+        curl \
+ && rm -rf /var/lib/apt/lists/*
+
+# sysand-amd64 / sysand-arm64 are populated via the publish-image workflow/job
+COPY --chmod=0755 sysand-${TARGETARCH}/sysand /usr/local/bin/sysand
```
simonas-drauksas-sensmetry, line 18: Should we add ENTRYPOINT?

https://www.docker.com/blog/docker-best-practices-choosing-between-run-cmd-and-entrypoint/

.gitignore
```diff
@@ -11,3 +11,8 @@ version.txt
 AGENTS.md
 AGENTS.local.md
 CLAUDE.md
+
+sysand-amd64/
+sysand-arm64/
+sysand-linux-arm64.tar.xz
+sysand-linux-x86_64.tar.xz
```
andrius-puksta-sensmetry: Better not add random stuff like this to gitignore.

Dockerfile
consideRatio: Its not obvious, i recall in a gitlab automation that I had to explicitly say I want to use a shell entrypoint so i could do misc things instead of ending up with the specific tool.

If we specify an entrypoint, I don't want to make the CI system unspecify it later - not 100% on the situation.

Dockerfile
```diff
@@ -0,0 +1,18 @@
+# syntax = docker/dockerfile:1.3
+
+FROM docker.io/library/ubuntu:24.04
```
andrius-puksta-sensmetry, line 3: `alpine` is much smaller. Is it unsuitable?

.gitignore
consideRatio: It is what you get if you manually copy paste steps from the workflow, which you may do to test it locally.

Dockerfile
consideRatio: We need glibc as i understand it for sysand, i could not for example run sysand on ruby:3.1 - default gitlab image.

I've not tested if it works on alpine directly, but I understood it as a requirement to have glibc and that its not on alpine.

consideRatio: (ubuntu is 29Mb - i explored these options quite a bit already)

.github/workflows/publish-image.yml
consideRatio: Maybe, but is now the time to refine this further?

.gitignore
consideRatio: I'll remove it, we don't have the time to care about this level of details =/

Dockerfile
simonas-drauksas-sensmetry: Why do we need shell access in GitLab? Couldn't the "publish" job in the CI take the artifacts from the previous job and just run `sysand publish`?

consideRatio: Okay, as long as that container runs with env vars for auth against the registry, setup when launching some other image with shell and curl.

Dockerfile
andrius-puksta-sensmetry: >We need glibc as i understand it for sysand

Indeed, I forgot.

PR #363, opened 2026-05-29
feat: support cloning local kpars

andrius-puksta-sensmetry:
Also improve a few error messages and improve spec compliance.

The large number of changed files is almost entirely removing `usage: []` and subsequently changed checksums.


consideRatio (approved):

consideRatio:
Depending on if this was functional in the past stable release, I figure it should be chanlogged under maintenance rather than new stuff. For now I added the new label.

andrius-puksta-sensmetry:
This is new, I guess I forgot to add it in original clone implementation.

PR #370, opened 2026-05-31
feat: add installation scripts

### LINKED ISSUE #367: Provide installer scripts (.sh / .ps1)
Can be used to reduce complexity of setup instructions etc.

Related a bit to #368.

consideRatio:
Add inspectable Unix and Windows install scripts, document cautious
and convenience install paths, and add a workflow that tests the
scripts against fake release archives.

Signed-off-by: Erik Sundell <erik.sundell@sensmetry.com>

---

I think it could be fine to put this higher up in the docs about installation, but figured its freshly introduced and not tested properly yet, so it could be better left below other install options for now.

I'm looking land this tomorrow and write trusted publishing docs with examples making use of it in CI context.

Instead of providing this example, it would become:

```yaml
      - name: Install sysand
        run: |
          ARCH=$(uname -m | sed 's/aarch64/arm64/')
          SYSAND_VERSION=$(
            curl -fsSL https://api.github.com/repos/sensmetry/sysand/releases \
              | awk -F'"' '/"tag_name":/ { tag=$4 } /"prerelease": true/ && !found { found=tag } END { if (found) print found }'
          )
          SYSAND_ARCHIVE="sysand-linux-${ARCH}.tar.xz"
          curl -fsSL "https://github.com/sensmetry/sysand/releases/download/${SYSAND_VERSION}/sysand-linux-${ARCH}.tar.xz" \
            | tar -xJ sysand
          install -m 0755 sysand /usr/local/bin/sysand
          sysand --version
```

```yaml
      - name: Install sysand
        run: |
          curl -fsSL https://client.sysand.com/install.sh | sh
```

- fixes #367

### DIFF
--- a/.github/workflows/test-install-script.yml
+++ b/.github/workflows/test-install-script.yml
@@ -1,0 +1,273 @@
+name: Test Install Script
+
+on:
+  push:
+    branches: [main]
+    tags: ["v*"]
+  pull_request:
+    branches: [main]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  plan:
+    runs-on: ubuntu-slim
+    outputs:
+      test: ${{ steps.decisions.outputs.test }}
+      gate: ${{ steps.decisions.outputs.gate }}
+    steps:
+      - uses: dorny/paths-filter@fbd0ab8f3e69293af611ebaee6363fc25e6d187d # v4.0.1
+        if: github.event_name == 'pull_request'
+        id: filter
+        with:
+          filters: |
+            src:
+              - 'docs/src/install.sh'
+              - 'docs/src/install.ps1'
+              - 'docs/src/getting_started/installation.md'
+              - '.github/workflows/test-install-script.yml'
+      - id: decisions
+        run: |
+          test=${{ steps.filter.outputs.src == 'true' || github.event_name == 'workflow_dispatch' || github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/') }}
+          gate=${{ github.event_name == 'pull_request' }}
+          echo "test=$test" >> "$GITHUB_OUTPUT"
+          echo "gate=$gate" >> "$GITHUB_OUTPUT"
+
+  unix:
+    name: Unix installer (${{ matrix.os }})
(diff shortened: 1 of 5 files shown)


consideRatio:
Rebased and force-pushed.

.github/workflows/test-install-script.yml
```diff
@@ -0,0 +1,273 @@
...
+            echo "expected --system-install followed by --install-dir to fail" >&2
+            exit 1
+          fi
+
+      - name: Stop fake release server
+        if: always()
+        run: |
+          if [ -f "$RUNNER_TEMP/http-server.pid" ]; then
+            kill "$(cat "$RUNNER_TEMP/http-server.pid")" || true
+          fi
```
andrius-puksta-sensmetry: Doesn't the server get killed when the last action is done?

docs/src/install.ps1
```diff
@@ -0,0 +1,127 @@
...
+if ($Help) {
+    Show-Usage
+    exit 0
+}
+
+if ([string]::IsNullOrWhiteSpace($Version)) {
+    Fail "Version must not be empty."
+}
+
+if ($Version -notmatch '^[A-Za-z0-9._-]+$') {
```
andrius-puksta-sensmetry: Underscore (`_`) is not allowed in SemVer.

.github/workflows/test-install-script.yml
consideRatio: Yeah, this is over engineered! Cleaning it up

docs/src/install.ps1
consideRatio: :+1: fixed, but still just relies on allowed characters rather than a regexp - as it must be doable from sh and powershell and then its messy to do regexp

docs/src/install.sh
```diff
@@ -0,0 +1,230 @@
...
+}
+
+run_system_install() {
+  if [ "$(id -u)" = "0" ]; then
+    "$@"
+  elif command -v sudo >/dev/null 2>&1; then
+    sudo "$@"
+  else
+    fail "sudo is required for --system-install when not running as root"
+  fi
```
andrius-puksta-sensmetry:
```suggestion
  else
    sudo "$@"
  fi
```
I think we can assume that any sysand-supported system will have `sudo`.

andrius-puksta-sensmetry (approved):

PR #372, opened 2026-06-01
feat: make built KPARs reproducible by setting OS/time metadata to fixed values

### LINKED ISSUE #343: Requirements for KPARs produced by Sysand
For zip metadata, so that produced KPARs are identical every time: 
* Set the OS to UNIX
* Set the last modified time to `zip::DateTime::None` (which means some specific date in 1980)

EDIT by Erik: note also [related issue in Gitlab](https://gitlab.com/sensmetry/internal2/tech/syside/sysand/home/-/work_items/1)

andrius-puksta-sensmetry:
Closes #343.

### DIFF
--- a/core/src/project/reqwest_kpar_download_tests.rs
+++ b/core/src/project/reqwest_kpar_download_tests.rs
@@ -31,6 +31,7 @@ fn basic_download_request() -> Result<(), Box<dyn std::error::Error>> {
 
         let options = zip::write::SimpleFileOptions::default()
             .compression_method(zip::CompressionMethod::Stored)
+            // TODO: why set permissions explicitly here?
             .unix_permissions(0o755);
 
         zip.start_file("some_root_dir/.project.json", options)?;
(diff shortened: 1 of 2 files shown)


core/src/project/reqwest_kpar_download_tests.rs
```diff
@@ -31,6 +31,7 @@ fn basic_download_request() -> Result<(), Box<dyn std::error::Error>> {
 
         let options = zip::write::SimpleFileOptions::default()
             .compression_method(zip::CompressionMethod::Stored)
+            // TODO: why set permissions explicitly here?
             .unix_permissions(0o755);
```
consideRatio, line 35: Not sure what the impact is, and why a project folder zip should have 755 which involves executable / `+x`

consideRatio (approved):

core/src/project/reqwest_kpar_download_tests.rs
andrius-puksta-sensmetry: Didn't want to change this for now, as this is only set in tests.

PR #380, opened 2026-06-03
docs: remove flakey workaround for the header's link targets

consideRatio:
React had server side rendered built pages, and were to hydrate retroactively. Then however it found that a header link had been changed, and had to rebuild the DOM, causing a flicker.

This removal fixes the thing you observed @andrius-puksta-sensmetry!


consideRatio:
Going for it, its one of these PRs I think is worth not having multiple people spend time on.

consideRatio:
Wasn't enough to fix it, but its the same issue.

PR #381, opened 2026-06-04
fix(sync): don't duplicate editable projects in env.toml that have no identifiers, include publisher in lockfile

andrius-puksta-sensmetry:
When `sysand sync` is run multiple times, editable projects present in `sysand-lock.toml` and not having an identifier will be included each time into `.sysand/env.toml`, causing duplication. Among such projects is generally the current project that the sync is run on.

Also fixes a bug that caused `publisher` to be omitted in `sysand-lock.toml` and by extension some `.sysand/env.toml` entries.

### DIFF
--- a/core/src/commands/lock.rs
+++ b/core/src/commands/lock.rs
@@ -163,16 +163,6 @@ pub fn do_lock_projects<
                     field: IncompleteField::Meta,
                 })
             })?;
-        // let canonical_digest = project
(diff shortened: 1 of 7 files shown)


consideRatio:
We haven't locked in what we write within the parenthesis like `fix(sync)` but perhaps individual commands could fit there - so instead of fix: sync: it would be fix(sync):

PR #382, opened 2026-06-04
feat(build): update index and checksum by default, provide opt out for index

### LINKED ISSUE #378: Dealing with outdated checksums
We need to decide how to deal with outdated checksums in `meta.checksum` when building. Currently we just include them as-is in the kpar, but then the index (correctly) rejects kpars with outdated checksums. Options:
1. always update to match file contents, inform if any updates were made
2. update if it's SHA256 (that's the only type we currently calculate, but KerML spec allows many others)
3. check and warn/inform if outdated
4. check and error out if outdated
5. do nothing (current situation)

My choice would be 1.

Similar situation is with the metamodel - index requires it to be set, but we don't say anything when building if it's not set. Since it's an optional field in the spec, not sure if it's appropriate to warn or not.

andrius-puksta-sensmetry:
This reverts the default flip done by #298. Index-only opt-out is provided, since it doesn't make sense to not update checksums to match actual file contents. Java bindings (and only these) will not update index, as the user can supply custom project metadata, including index, and updating the index silently on build would override the supplied one.

Now the behaviour matches user expectations in almost all cases, and requires less friction to build KPARs publishable to the index.

In bindings, also allow path usages in the KPARs being built, since this is not yet configurable by the caller.

Fixes #378

### DIFF
--- a/sysand/src/lib.rs
+++ b/sysand/src/lib.rs
@@ -414,6 +414,7 @@ pub fn run_cli(args: cli::Args) -> Result<()> {
             }
         }
         Command::Sync { resolution_opts } => {
+            // TODO: only print this if we actually skip install of any std libs
             let provided_iris = if !resolution_opts.include_std {
                 crate::logger::warn_std_deps();
                 known_std_libs()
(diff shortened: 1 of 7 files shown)


sysand/src/lib.rs
```diff
@@ -667,8 +668,9 @@ pub fn run_cli(args: cli::Args) -> Result<()> {
         Command::Build {
             path,
             compression,
-            update_meta,
+            update_meta: _,
```
vytautas-astrauskas-sensmetry: Instead of silently ignoring `--update-meta`, I think it would be better to show a warning that the flag is ignored.

vytautas-astrauskas-sensmetry (approved):
Besides silently ignoring flags, looks good to me.

PR #390, opened 2026-06-09
chore: improve robustness of install script

consideRatio:
Not ready for review, testing CI pipelines and trying to pin down what makes sense doing etc still.

### DIFF
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -1,12 +1,13 @@
 # Sysand installer for Windows.
 #
 # This script is intentionally small and direct so it can be inspected before
-# running. It downloads a release archive, extracts sysand.exe, and copies it
-# into an install directory.
+# running. It downloads a release archive, extracts sysand.exe, and replaces the
+# installed binary.
 
 param(
     [string]$Version = "latest",
-    [string]$InstallDir = "$env:LOCALAPPDATA\Programs\Sysand",
+    [string]$InstallDir = "$env:LOCALAPPDATA\Programs\Sensmetry\Sysand\bin",
+    [switch]$NoModifyPath,
     [switch]$Help
 )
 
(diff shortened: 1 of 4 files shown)


scripts/install.ps1
```diff
@@ -37,15 +40,115 @@ function Fail {
...
+Detected PowerShell $($PSVersionTable.PSVersion).
+"@
+    }
+
+    if ([System.Enum]::GetNames([System.Net.SecurityProtocolType]) -notcontains "Tls12") {
+        Fail @"
+Installing Sysand requires TLS 1.2 support from .NET Framework 4.5 or later.
+Install a newer .NET Framework before running this installer.
+"@
+    }
```
andrius-puksta-sensmetry, line 56: These checks only make sense if we target ancient systems. Since Rust requires Windows 10, PowerShell >=5 is guaranteed, and so is .NET framework >=4.5.

andrius-puksta-sensmetry: (to be precise, powershell 5 was released with win10 1607 in 2016, which is not exactly baseline win10, but close enough to make it worth getting rid of extra checks)

andrius-puksta-sensmetry: Windows versions support: https://endoflife.date/windows

andrius-puksta-sensmetry: Rust requirements: https://doc.rust-lang.org/nightly/rustc/platform-support.html#tier-1-with-host-tools

andrius-puksta-sensmetry: .NET framework 4.6 was also released with win10 1607: https://endoflife.date/dotnetfx

consideRatio:
I'll make sure to lean on win10+ without more details checked @andrius-puksta-sensmetry! I'll close this PR and open a new when its ready!

PR #391, opened 2026-06-09
feat: add linux musl build, make GNU build require only glibc 2.35+

### LINKED ISSUE #373: Provided build without glibc dependency (static build)?
We could make sysand work on alpine etc, if we some dependencies on glibc i think. I'm not sure how to do it etc, or if its doable, but I'm used to installing special binaries on alpine images compared to debian based images.  See for example https://github.com/krallin/tini#statically-linked-version.

If we provide this, we could then also have the installer script automatically install the relevant version depending on its detected environment.

consideRatio:
When we build on linux we used ubuntu 24.04 runners without declaring an explicit "target" for the rust build. That made us produce a binary that dynamically link / references glibc 2.39+ and hence requires that on the linux machine running the binary.

With this PR, we start building on ubuntu 22.04 instead, lowering the requirement of glibc from 2.39+ to 2.35+. We also add CI jobs to test and build on ubuntu 24.04 but providing an explicit `musl` target, which makes the built binary no longer reference glibc. A `musl` binary could work on all linux distros, but there may be differences. @tilowiklundSensmetry recommended we provide both a gnu/glibc build and musl build, which is what we now do.

---

This PR is paired with https://gitlab.com/sensmetry/internal2/tech/syside/sysand/sysand-signing/-/merge_requests/14 which ensures that the nighly builds produced are signed, packaged, and uploaded back in a GitHub Release.

The binary size of musl vs gnu version  was the same on linux arm (8.36MB), but was slightly larger on linux amd (8.39MB to 8.63MB).

The installer script should also be updated, but I'd like to do that in a separate PR if/when this is merged.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -96,24 +96,64 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os:
-          - ubuntu-24.04
-          - ubuntu-24.04-arm
-          - windows-latest
-          - windows-11-arm
-          - macos-15-intel # x86
-          - macos-latest # arm64
+        include:
+          # About Linux GNU vs MUSL builds:
+          #
+          # GNU:  debian, ubuntu, rhel, fedora, arch, ...
+          # MUSL: scratch, alpine, gentoo, ...
+          #
+          # ubuntu-22.04 is used intentionally instead of ubuntu-24.04 for the
+          # libc=gnu builds. With ubuntu 22.04 instead of 24.04, we build a
+          # binary dynamically linking against glibc 2.35+ instead of 2.39+.
+          # That makes the build compatible with more versions of various glibc
+          # based distributions, like ubuntu 22.04+, debian 12+, and rhel 10+.
+          #
+          # The MUSL builds are static, with no dynamically linked dependencies,
+          # and can work on GNU / glibc based distributions, but not the other
+          # way around. GNU / MUSL may have differences such as DNS resolution
+          # behavior in complicated cases, but is mostly expected to work the
+          # same way.
+          #
+          - os: ubuntu-22.04
+            libc: gnu
+          - os: ubuntu-22.04-arm
+            libc: gnu
+          - os: ubuntu-24.04
+            target: x86_64-unknown-linux-musl
+            libc: musl
+          - os: ubuntu-24.04-arm
+            target: aarch64-unknown-linux-musl
+            libc: musl
+          - os: windows-latest
+            target: ""
+          - os: windows-11-arm
+            target: ""
+          - os: macos-15-intel # x86
+            target: ""
+          - os: macos-latest # arm64
+            target: ""
 
     steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/rust.yml
```diff
@@ -96,24 +96,44 @@ jobs:
...
+          echo "CC_${target//-/_}=musl-gcc" >> "$GITHUB_ENV"
+          echo "CARGO_BUILD_TARGET=$target" >> "$GITHUB_ENV"
+
       - name: cargo test ... (Rust library & CLI)
         run: >-
           cargo test --locked
           --package sysand-core
           --package sysand
-          --features sysand-core/filesystem,sysand-core/networking,sysand-core/js,sysand-core/python,alltests,kpar-bzip2,kpar-zstd,kpar-xz,kpar-ppmd
+          --features sysand-core/filesystem,sysand-core/networking,alltests,kpar-bzip2,kpar-zstd,kpar-xz,kpar-ppmd
```
consideRatio, line 156: I could not get this to build with `sysand-core/python` feature enabled. However, when we build the wheels for python, we provide specific builds for `manylinux` (debian, ubuntu, fedora, rhel, ...) and `musllinux` (alpine, ...) and both works.

consideRatio: I removed both sysand-core/js and sysand-core/python since we cover those in dedicated jobs anyhow.

andrius-puksta-sensmetry: Well, `js` and `python` features don't enable any tests in core anyhow.

.github/workflows/rust.yml
```diff
@@ -96,60 +96,168 @@ jobs:
...
       - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
 
+      # Ubuntu 22.04 provides glibc 2.35. Some native C dependencies can
+      # otherwise pick a newer compiler default and emit references to C23
+      # glibc symbols that are unavailable in the target runtime.
+      - name: (linux gnu) Configure Linux GNU C standard
+        if: matrix.libc == 'gnu'
+        run: |
+          echo "CFLAGS=-std=gnu17" >> "$GITHUB_ENV"
+          echo "CXXFLAGS=-std=gnu++17" >> "$GITHUB_ENV"
```
andrius-puksta-sensmetry: I don't get this. If we build on ubuntu 22.04, the binary will be compatible with glibc 2.35 and up.

andrius-puksta-sensmetry: If this is about not finding the symbols when building, let's address these if/when they arise.

consideRatio: It takes a while to iterate, and maybe this wasnt a true need but a not reset cache causing it to be needed. Im not 100%.

I can iterate and delete this and clear cache, if it turns out its needed, we retain it and go, otherwise delete - does that sound OK?

consideRatio: Wieee success it worked, nice! I really appreciate avoiding complexity like this!

andrius-puksta-sensmetry (approved):
Good that sanity checks for linking to musl/glibc are included.

PR #393, opened 2026-06-10
feat: publish an alpine based image variant with -alpine suffixed tags

### LINKED ISSUE #392: Provide `-alpine` option to `ubuntu` based CI image
alpine is ~4 MB, ubuntu is ~28 MB. By providing a -alpine version of our CI image, it can be even slimmer.

consideRatio:
The alpine image uses the static musl release binaries, so the
sysand-signing project must package sysand-linux-{x86_64,arm64}-musl.tar.gz
release assets, each containing a single binary named sysand at the
tarball root, matching the -gnu tarball layout.

Signed-off-by: Erik Sundell <erik.sundell+2025@sensmetry.com>

### DIFF
--- a/.github/workflows/publish-images.yml
+++ b/.github/workflows/publish-images.yml
@@ -1,4 +1,4 @@
-name: Publish Image
+name: Publish Images
 
 on:
   release:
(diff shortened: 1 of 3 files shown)


andrius-puksta-sensmetry (approved):
Not an expert in these things, so hopefully it works.

PR #394, opened 2026-06-10
feat!: rework installation scripts to improve UX and support alpine etc

### LINKED ISSUE #385: Installer script improvements
## High level goals

In gist, I'd like us to have as simple and clear installation instructions as https://docs.sysand.com/client/getting-started/installation/, but it requires some work on the installer script itself.

- [ ] Ease of use
  - [ ] Handles PATH complexity by default
  - [ ] Handles minimum requirements, errors clearly
  - [ ] Function for initial install, but also for updates
- [ ] Verified to work
  - [ ] Install verified for typical Windows user
  - [ ] Install verified for typical Mac user
  - [ ] Install verified for typical Linux user
  - [ ] Install verified for use in GitHub Actions

## Decisions

- Scripts must be written in a way that makes it very readable for security audit purposes
- Support installs and updates
- Make the install / update operation atomic
- Only user scoped installations
  - `$HOME/.local/bin` and `%LOCALAPPDATA%\Programs\Sensmetry\Sysand\bin` respectively
- PATH handling:
  - if GITHUB_PATH is detected, we configure it, log doing so, and moves on
  - Should be idempotent (disregard GITHUB_PATH)
  - Ensure binary is on PATH for future sessions (log doing so or failing to do so, support default shells on ubuntu debian mac at least)
  - Print instructions on how to update PATH for current session (don't assume .ps1 script was run from powershell, print instructions for both powershell and cmd.exe)
- System pre-install checks:
  - LINUX: if glibc is found and 2.35+ we install the gnu version, otherwise the musl version
  - MAC: we have requirements for the produced build artifact, we should check it
  - WINDOWS: we have requirements for win10+ and powershell 5+ we should check it
- A piped install.sh can run and abort mid-way, ensure it loads fully before running

consideRatio:
## Breaking changes

- The installation script doesn't accept flags any more, only environment variables.
- The installation is only compatible with the version 0.1.2 and later

---

Configure both installers via SYSAND_VERSION instead of flags so the same
mechanism works for curl|sh and irm|iex; removed flags fail with migration
errors. Installs are user-scoped only: $HOME/.local/bin on Unix and
%LOCALAPPDATA%\Programs\Sensmetry\Sysand\bin on Windows, where the legacy
Programs\Sysand binary and PATH entry are cleaned up.

Install and update atomically: Unix stages in the destination directory and
renames into place; Windows moves a possibly running sysand.exe aside before
renaming the staged binary in, with rollback and leftover cleanup.

Select the Linux artifact by detected glibc (2.35+ gets gnu, anything else
the static musl build), require macOS 11 on arm64 and 10.12 on x86_64 (the
Rust toolchain minimums), prefer the native build under Rosetta 2, and
require Windows 10.

Handle PATH in three tiers: append to GITHUB_PATH when set, skip persistence
in other CI (CI env var) where the job shell is the only session, otherwise
persist idempotently (bash/zsh profiles on Unix, the HKCU Environment Path
registry value on Windows, preserving REG_EXPAND_SZ entries) and print
current-session instructions for sh, PowerShell, and cmd.exe.

The sh installer runs everything from a main function called on the last
line, so a truncated curl|sh download executes nothing.

The test workflow covers the new behavior, adding alpine and debian:11
container jobs for musl selection and a Windows PowerShell 5.1 leg alongside
pwsh.

---

- Fixes https://github.com/sensmetry/sysand/issues/385
- Changes introduced to be removed later tracked by https://github.com/sensmetry/sysand/issues/395
- This PR was developed with Fable 5, I'm happy with its job

### DIFF
--- a/.github/workflows/installation-script.yml
+++ b/.github/workflows/installation-script.yml
@@ -49,12 +49,20 @@ jobs:
         include:
           - os: ubuntu-24.04
             asset: sysand-linux-x86_64-gnu.tar.gz
+            shell: /bin/bash
+            profile: .bashrc
           - os: ubuntu-24.04-arm
             asset: sysand-linux-arm64-gnu.tar.gz
+            shell: /bin/bash
+            profile: .bashrc
           - os: macos-15-intel
             asset: sysand-macos-x86_64.tar.gz
+            shell: /bin/zsh
+            profile: .zshrc
           - os: macos-latest
             asset: sysand-macos-arm64.tar.gz
+            shell: /bin/zsh
+            profile: .zshrc
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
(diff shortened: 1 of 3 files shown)


consideRatio:
Merging this doesn't mean sysand.com updates its script. We should do that only after we have a new release out.

scripts/install.sh
```diff
@@ -2,38 +2,39 @@
...
-                        Default: $HOME/.local/bin
+Installs the sysand binary to $HOME/.local/bin, replacing any existing
+installation, and ensures that directory is on PATH.
 
-  --system-install      Install into /usr/local/bin.
-                        If the current user is not root, the install step uses
-                        sudo.
+Configuration via environment variables:
+  SYSAND_VERSION   Release version to install, for example 0.1.0 or
+                   0.1.0-rc.1. A leading "v" is also accepted.
```
andrius-puksta-sensmetry:
```suggestion
  SYSAND_VERSION   Release version to install, for example 0.1.2 or
                   0.1.2-dev.1. A leading "v" is also accepted.
```
Let's not suggest versions which won't work.

andrius-puksta-sensmetry: (for consistency, also update install.ps1)

scripts/install.sh
```diff
@@ -152,78 +216,173 @@ build_download_url() {
   fi
 }
 
+download_failed() {
+  if [ "${libc:-}" = "musl" ]; then
+    echo "note: musl assets are only published for sysand releases newer than 0.1.1" >&2
+  fi
```
andrius-puksta-sensmetry: The same applies to glibc builds.

scripts/install.sh
```diff
@@ -152,78 +216,173 @@ build_download_url() {
...
-need_cmd uname
-need_cmd mktemp
-need_cmd tar
+main() {
+  parse_args "$@"
+
+  need_cmd uname
+  need_cmd mktemp
+  need_cmd tar
+  [ -n "${HOME:-}" ] || fail "HOME must be set"
```
andrius-puksta-sensmetry, line 366: Base utils and HOME can be assumed to be present; if not, the user is probably an expert anyway.
```suggestion
```
This also allows removing `need_cmd`.

.github/workflows/installation-script.yml
```diff
@@ -233,22 +412,206 @@ jobs:
...
+          }
+          $backup = Import-Clixml -Path $backupFile
+          $envKey = [Microsoft.Win32.Registry]::CurrentUser.OpenSubKey("Environment", $true)
+          if ($backup.HasPath) {
+              $kind = [Enum]::Parse([Microsoft.Win32.RegistryValueKind], $backup.Kind)
+              $envKey.SetValue("Path", $backup.RawPath, $kind)
+          } elseif ($envKey.GetValueNames() -contains "Path") {
+              $envKey.DeleteValue("Path")
+          }
+          $envKey.Dispose()
```
andrius-puksta-sensmetry: This is the last step, cleanup not needed (also backup not needed above)

andrius-puksta-sensmetry (commented):
Just a few nits. Reviewing scripts and yaml is really exhausting...

scripts/install.sh
consideRatio: Let's retain them all.
- mktemp isn't POSIX defined while its around almost everywhere.
- tar isn't around in alpine for example
- HOME can very well end up being blank string or unset for some reason and if its a blank string we modify a weird path by mistake, its not as bad as `rm -rf /$MY_PATH` and then MY_PATH was a blank string but still.

Maybe uname is the safest to delete, but still, its nice to simply know what things are being run.

Removing these lines reduces lines of code, but it doesn't really reduce complexity  because then suddenly we must consider all situations when they could potentially be missing etc. I'd rather feel very confident knowing that users error clearly, and whenver people report an error in this script I can be confident they do have `tar` installed etc and not error cryptically without me realizing its actually `tar` missing or HOME being a blank string.

scripts/install.sh
andrius-puksta-sensmetry: Why is this marked resolved?

andrius-puksta-sensmetry (approved):
Aside from the nit above this is good to go.

consideRatio:
@andrius-puksta-sensmetry sorry for the trouble, i ended up thinking i had pushed a commit fixing things, but hadn't then i saw that and "fixed it" but never did it correctly. NOW though I have pushed the commit of relevance: 
cd754e8.

(I had the wrong remote configured when pushing my branch, so it looked OK but wasn't ending up updating the PR)

scripts/install.sh
andrius-puksta-sensmetry: We explicitly reject versions `< 0.1.2`, so suggesting them here is counterproductive.

PR #399, opened 2026-06-12
fix: don't hardcode file name in error

andrius-puksta-sensmetry:
Also give nice error if project is not at the root of the kpar in `index add`.


consideRatio (approved):

consideRatio:
I figure it belongs as bug label more than enhancement in the changelog, adjust label?

PR #400, opened 2026-06-12
feat: perform most of KPAR pre-publish checks before publishing

### LINKED ISSUE #389: User-friendly license setting
Sysand Index requires the license(s) specified in `project.license` to be present as `.txt` files in `LICENSES/`. In most cases, these files will just contain canonical license texts from SPDX. [`spdx` crate](https://docs.rs/spdx/latest/spdx/) that we use on the client side conveniently can provide canonical SPDX license texts. Providing these manually is tedious, and users may not know where to obtain canonical texts.

My proposal: `sysand info license --set <expression>` should automatically create the required files by default. Specifically:

- it must be default, but provide an opt-out
- license files will be named according to our requirements
- each license file will include full canonical text of the license it represents
- existing files won't be touched, e.g. `sysand info license --set MIT` won't touch `LICENSES/MIT.txt` if present
- a warning will be generated if the folder contains unneeded files (e.g. previous licenses that are no longer needed by the license expression)

andrius-puksta-sensmetry:
- checks may be a bit more/less strict than on the server
- errors are currently descriptive, but most are not actionable. Not sure where to provide suggestions (this also applies to many other errors in core that are user-fixable):
  - core should not mention CLI flags, since it's also used by bindings
  - unclear how to provide them in CLI either: matching each error variant and adding more details to some of them is tedious and very verbose
- fix publish tests to now produce kpars that pass our checks

Closes #389 - I decided against compiling in all license texts, since that would add a few extra MB to the binary, which is not justifiable for such a rarely needed feature. Instead a detailed help message has been added when failing due to missing license in publish, and a similar one is printed when setting the license with `sysand info license --set`

### DIFF
--- a/core/src/project/local_kpar.rs
+++ b/core/src/project/local_kpar.rs
@@ -460,10 +460,6 @@ impl LocalKParProjectRaw {
                 .map_err(|e| FsIoError::WriteFile(path.as_ref().into(), e))?;
         }
 
-        // This can only fail if the comment is too long
-        zip.set_comment(concat!("produced by: sysand v", env!("CARGO_PKG_VERSION")))
-            .unwrap();
-
         zip.finish()
             .map_err(|e| ZipArchiveError::Finish(path.as_ref().into(), e))?;
 
--- a/core/src/project/local_kpar.rs
+++ b/core/src/project/local_kpar.rs
@@ -460,6 +460,10 @@ impl LocalKParProjectRaw {
                 .map_err(|e| FsIoError::WriteFile(path.as_ref().into(), e))?;
         }
 
+        // This can only fail if the comment is too long
+        zip.set_comment(concat!("produced by: sysand v", env!("CARGO_PKG_VERSION")))
+            .unwrap();
+
         zip.finish()
             .map_err(|e| ZipArchiveError::Finish(path.as_ref().into(), e))?;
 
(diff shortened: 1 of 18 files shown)


core/src/project/local_kpar.rs
```diff
@@ -460,6 +460,10 @@ impl LocalKParProjectRaw {
                 .map_err(|e| FsIoError::WriteFile(path.as_ref().into(), e))?;
         }
 
+        // This can only fail if the comment is too long
+        zip.set_comment(concat!("produced by: sysand v", env!("CARGO_PKG_VERSION")))
+            .unwrap();
```
andrius-puksta-sensmetry: This is a random idea I had. Potential uses:
- skip most publish checks, since our kpars always satisfy them
- if/when we have some sort of telemetry, this allows us to estimate the proportion of kpars built with sysand and thus sysand's usage

andrius-puksta-sensmetry: Not sure if we should keep this.

andrius-puksta-sensmetry: > if/when we have some sort of telemetry, this allows us to estimate the proportion of kpars built with sysand and thus sysand's usage

On the other hand, if sysand is already being used, it's quite likely kpars will be built by it...

Removing this, no good use case. First use can better be solved by e.g. assuming that if the kpar is in our default lcoation, it's built by sysand. Also, it would probably be a good idea to allow building within publish command, at least as an option.

consideRatio (approved):

andrius-puksta-sensmetry:
Found a few bugs.

PR #401, opened 2026-06-12
release: prepare 0.1.3

andrius-puksta-sensmetry:


andrius-puksta-sensmetry:
Let's go for the release now, the bugfix is more important than waiting for #400.

PR #406, opened 2026-06-15
feat: improve performance of include, build and exclude

andrius-puksta-sensmetry:
`build` times for a large test project went from ~5 min to ~3 s.
Similar results for `include`, they `build` does very similar work to `include`

Also fix `build` producing truncated `.meta.json` files for large projects by using `write_all()` instead of `write()`.

Previously, during build, every file in the project was copied multiple times:
1. make a temporary copy of the whole project (needed to avoid touching original, because `.meta.json` was modified in place)
2. for each source file:
    1. read the file to extract symbols
    2. update (read+write) `.meta.json`
    3. read the file to calculate checksum
    4. update (read+write) `.meta.json`
    5. read the file to put into the kpar

Include performs most of the same stuff, but it does not make a temp copy of the project and obviously does not create a kpar.

### DIFF
--- a/bindings/java/src/lib.rs
+++ b/bindings/java/src/lib.rs
@@ -410,18 +410,6 @@ fn handle_build_error(env: &mut JNIEnv<'_>, error: KParBuildError<LocalSrcError>
                 format!("Project read error: {}", error),
             );
         }
-        KParBuildError::LocalSrc(error) => {
-            env.throw_exception(
-                ExceptionKind::SysandException,
-                format!("Local src error: {}", error),
-            );
-        }
-        KParBuildError::IncompleteSource(error) => {
-            env.throw_exception(
-                ExceptionKind::SysandException,
-                format!("Incomplete source error: {}", error),
-            );
-        }
         KParBuildError::Io(error) => {
             env.throw_exception(
                 ExceptionKind::SysandException,
@@ -454,6 +442,12 @@ fn handle_build_error(env: &mut JNIEnv<'_>, error: KParBuildError<LocalSrcError>
         }
         KParBuildError::MissingMeta => {
             env.throw_exception(ExceptionKind::SysandException, "Missing project metadata");
+        }
+        KParBuildError::MissingInfoMeta => {
+            env.throw_exception(
+                ExceptionKind::SysandException,
+                "Missing project information and metadata",
+            );
         }
         KParBuildError::Zip(error) => {
             env.throw_exception(
--- a/bindings/py/src/lib.rs
+++ b/bindings/py/src/lib.rs
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 // SPDX-FileCopyrightText: © 2025 Sysand contributors <opensource@sensmetry.com>
 
-use std::{collections::HashMap, process::ExitCode, sync::Arc};
+use std::{collections::HashMap, iter, process::ExitCode, sync::Arc};
 
 use camino::{Utf8Path, Utf8PathBuf};
 use pyo3::{
@@ -41,6 +41,7 @@ use sysand_core::{
     stdlib::known_std_libs,
     symbols::Language,
 };
+use typed_path::Utf8UnixPathBuf;
 
 #[pyfunction(name = "_run_cli")]
 fn run_cli(args: Vec<String>) -> PyResult<bool> {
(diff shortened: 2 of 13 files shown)


consideRatio:
## Codex GPT 5.5

### Prompt

```
› Can you review the branch local PR thoroughly? I'm providing PR title/description below as well.

  I'm looking for essential issues that should block a merge, but we need to get a merge soon so can't bother with non essential details. If you identify something worth presenting, confirm it and ensure present a text written as a staff technical writer about
  it.

  ## PR title

  feat: improve performance of include, build and exclude

  ## PR description

  `build` times for a large test project went from ~5 min to ~3 s.
  Similar results for `include`, they `build` does very similar work to `include`

  This should also fix `build` producing truncated `.meta.json` files for large projects, since it's now written directly into the archive, and so write interleaving/writeback caches are not a concern.

  Previously, during build, every file in the project was copied multiple times:
  1. make a temporary copy of the whole project (needed to avoid touching original, because `.meta.json` was modified in place)
  2. for each source file:
      1. read the file to extract symbols
      2. update (read+write) `.meta.json`
      3. read the file to calculate checksum
      4. update (read+write) `.meta.json`
      5. read the file to put into the kpar

  Include performs most of the same stuff, but it does not make a temp copy of the project and obviously does not create a kpar.
```

### Findings

```
  1. core/src/commands/build.rs:260 packages unvalidated raw metadata now. The old path called meta.validate() before selecting source files and writing the archive; this branch goes from raw get_project() metadata to serialization at core/src/commands/
     build.rs:438. I confirmed sysand build now succeeds and embeds "created":"not-a-date" in .meta.json. That should block merge because build can emit an invalid KPAR that later commands reject.

  2. core/src/commands/include.rs:72 makes --no-index-symbols destructive. The new implementation always removes existing symbols for the included file, then only re-adds them when index_symbols is true at core/src/commands/include.rs:81. I confirmed the
     regression: include once creates {"P":"test.sysml"}, then sysand include --no-index-symbols test.sysml changes the index to {}. The flag says not to detect/add symbols; it should not erase an existing index.

  3. core/src/commands/build.rs:442 still writes .project.json and .meta.json with write and ignores the returned byte count. ZipWriter::write delegates to the inner writer and can legally accept only a prefix, so this code path can still create truncated
     manifest entries. Given the PR explicitly targets truncated .meta.json output, these two calls should use write_all.

```

consideRatio:
## Claude Opus 4.8 review

### Prompt

Same as codex above.

### Report

● I've completed a thorough review of the branch. The refactor is sound overall — it compiles cleanly, and the rewritten include/exclude/build flows are behaviorally equivalent to the trait helpers they replace (single get_meta/put_meta round-trip instead of per-file read-modify-write, checksums and symbols derived from one in-memory read of each source). I found two issues worth your attention before merge, one of which directly undercuts a stated goal of the PR.

  ---
  Blocking: .project.json and .meta.json are written with write, not write_all

  core/src/commands/build.rs:442 and :447

  ```rust
  zip.write(info_content.as_bytes())   // .project.json
  zip.write(meta_content.as_bytes())   // .meta.json
  ```

  Every source file and extra file in the same function is written with write_all (lines 350, 391, 428), but the two metadata entries use write, and the returned byte count is discarded.

  io::Write::write is permitted to perform a short write — it returns the number of bytes actually accepted, which may be fewer than the buffer length. I traced this through the zip 8.6.0 ZipWriter::write implementation (src/write.rs:762): it forwards the call straight to the inner writer and returns that writer's count verbatim. For the Stored compression method (reachable via the store option) the inner writer is the raw archive File, so a short write propagates directly and silently truncates the entry. The default Deflated method happens to mask this — the flate2 encoder consumes the whole input buffer — which is precisely why a truncation here would pass casual testing and the existing test suite, then surface only on the --store path or under load.

  This is the same failure class the PR sets out to eliminate ("fix build producing truncated .meta.json files for large projects"), and .meta.json is the entry that grows without bound for large projects. Leaving these two calls as write reintroduces the exact risk the change is meant to close, on the one entry that matters most. The fix is a one-line change to write_all for both, matching the surrounding source-file writes.

  Note this was carried over verbatim from the deleted LocalKParProjectRaw::from_project, so it is a pre-existing latent bug rather than newly introduced — but the PR moves this code and is explicitly about write integrity, so it is the right place to correct it.

  ---
  Minor correctness: wrong error variant when .meta.json is missing

  core/src/commands/build.rs:264

  ```rust
  (None, Some(_)) => return Err(KParBuildError::MissingInfo),
  (Some(_), None) => return Err(KParBuildError::MissingInfo),   // should be MissingMeta
  ```

  The second arm is reached when project info is present but metadata is absent, yet it returns MissingInfo, whose message reads "missing project info file .project.json". A user who is actually missing .meta.json gets pointed at the wrong file.
  The dedicated MissingMeta variant ("missing project metadata file .meta.json") already exists and is the correct return here. One-token fix while the file is open.

  ---
  Everything else I checked held up: the validate() call dropped from build is safe because the new code reconstructs meta such that its invariant (every indexed file has a checksum) holds by construction — the index is rebuilt only from files that are also given checksums; the error cleanup that removes a partial archive is correctly ordered after the ZipWriter is dropped; and the do_include/do_exclude signature changes to iterators are threaded correctly through the CLI and both bindings.

consideRatio (approved):
I don't want this to be gated behind further review by me. I didn't spot issues myself, but codex/opus found some things that seems reasonable to look at.

PR #411, opened 2026-06-16
bindings: small improvements to Python API

simonas-drauksas-sensmetry:
This PR contains small Python API improvements that I need for integrating Sysand into Syside ReqIF functionality, namely:
1. The ability to obtain the project root (Sysand CLI already can do this through `print-root`, but not in Python)
2. The ability to obtain a list of dependencies the current project depends on (so I could check if the ReqIF library is imported)

These changes required a bit of changes on the Sysand core / CLI, namely the introduction of "source modes":
- `own`, which represents the "I only want to list the `included` files" and was achievable through CLI as `sysand sources --no-deps` (still possible)
- `own_with_deps`, which is what the default is `sysand sources`
- `deps`, which is new and returns only the dependencies without `included` files

While the CLI only got a new flag `--only-deps` and thus there is no breaking change on the CLI side, the Python API lost the `include_deps` argument and gained `no_deps` and `only_deps` instead to better align with the CLI. Not sure if we should consider this a breaking change, or should I try to keep what was already in the Python API?

For context, this is how I would need to get only the deps with the current API:

```python
all_srcs = sysand.sources(root, env_path=root / ".sysand")
own_srcs = sysand.sources(root, env_path=root / ".sysand", include_deps=False)
dep_srcs = [p for p in all_srcs if p not in set(own_srcs)]
```

While after this PR I will be able to do:

```python
dep_srcs = sysand.sources(root, env_path=root / ".sysand", only_deps=True)
```

P.S.: The commits are pretty atomic and contain nice commit messages. If not a lot of changes after review are required, I'd like NOT to squash the commits on merge.

### DIFF
--- a/bindings/py/python/sysand/_model.py
+++ b/bindings/py/python/sysand/_model.py
@@ -1,5 +1,5 @@
 # SPDX-License-Identifier: MIT OR Apache-2.0
-# SPDX-FileCopyrightText: © 2025 Sysand contributors <opensource@sensmetry.com>
+# SPDX-FileCopyrightText: © 2025-2026 Sysand contributors <opensource@sensmetry.com>
 
 from enum import Enum, auto
 import typing
@@ -36,6 +36,17 @@ class InterchangeProjectMetadata(typing.TypedDict):
     checksum: typing.Optional[typing.List[InterchangeProjectChecksum]]
 
 
+class Dependencies(Enum):
+    NONE = auto()
+    """Do not list any dependency sources"""
+    DEPS = auto()
+    """List dependency sources, excluding standard libraries"""
+    DEPS_STD = auto()
+    """List dependency sources, including standard libraries"""
+    STD = auto()
+    """List only standard-library dependency sources"""
+
+
 class CompressionMethod(Enum):
     STORED = auto()
     """Store the files as is"""
@@ -56,5 +67,6 @@ __all__ = [
     "InterchangeProjectInfo",
     "InterchangeProjectChecksum",
     "InterchangeProjectMetadata",
+    "Dependencies",
     "CompressionMethod",
 ]
--- a/bindings/py/src/lib.rs
+++ b/bindings/py/src/lib.rs
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 // SPDX-FileCopyrightText: © 2025 Sysand contributors <opensource@sensmetry.com>
 
-use std::{collections::HashMap, iter, process::ExitCode, sync::Arc};
+use std::{iter, process::ExitCode, sync::Arc};
 
 use camino::{Utf8Path, Utf8PathBuf};
 use pyo3::{
@@ -28,7 +28,7 @@ use sysand_core::{
     include::do_include,
     info::{InfoError, InfoProjectError, do_info, do_info_project},
     init::InitError,
-    model::{InterchangeProjectInfoRaw, InterchangeProjectMetadataRaw},
+    model::{InterchangeProjectInfoRaw, InterchangeProjectMetadataRaw, InterchangeProjectUsage},
     project::{
         ProjectRead as _,
         local_kpar::{KparInnerPath, LocalKParProject},
@@ -38,8 +38,7 @@ use sysand_core::{
     remove::do_remove_guess,
     resolve::{net_utils::create_reqwest_client, standard::standard_resolver},
     root::do_root,
-    sources::{do_sources_local_src_project_no_deps, find_project_dependencies},
-    stdlib::known_std_libs,
+    sources::{Dependencies, do_sources_local_src_project_no_deps, resolve_dependencies},
     symbols::Language,
     utils::format_err,
 };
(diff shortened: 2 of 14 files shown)


andrius-puksta-sensmetry:
> Not sure if we should consider this a breaking change, or should I try to keep what was already in the Python API?

For now I don't consider bindings/core API a part of stability guarantee; too few users to matter.

andrius-puksta-sensmetry:
>  gained no_deps and only_deps instead to better align with the CLI

It would be better to use an enum in Python API, otherwise we have to check if both `no_deps` and `only_deps` are given. IMO CLI should also change, but since this is breaking, CLI change won't happen until 0.2.0.

bindings/py/python/sysand/_model.py
```diff
@@ -36,6 +36,34 @@ class InterchangeProjectMetadata(typing.TypedDict):
     checksum: typing.Optional[typing.List[InterchangeProjectChecksum]]
 
 
+class SourcesMode(Enum):
+    OWN = auto()
+    """Only the project's own sources"""
+    OWN_WITH_DEPS = auto()
+    """The project's own sources together with its dependencies' sources"""
+    DEPS = auto()
+    """Only the dependencies' sources"""
```
andrius-puksta-sensmetry: IMO a better design would be to have:
- a flag/arg `no_own` to exclude self sources, which are included by default
- a flag/arg `include` with variants:
  - `deps` (show deps excluding std)
  - `deps-std` (show deps including std)
  - `std` (show only std)
  - `none` (don't show deps)

Any combination of `no_own` and `include` is valid.

andrius-puksta-sensmetry: (with no dependencies shown by default)

bindings/py/src/lib.rs
```diff
@@ -250,22 +262,18 @@ fn do_build_py(
...
     include_std: bool,
 ) -> PyResult<Vec<String>> {
     let _ = pyo3_log::try_init();
 
-    let provided_iris = if !include_std {
-        known_std_libs()
-    } else {
-        HashMap::default()
-    };
+    let mode = SourcesMode::try_from(mode).map_err(|e| PyValueError::new_err(e.to_string()))?;
```
andrius-puksta-sensmetry: This does not check for conflicts with `include_std`.

core/src/commands/root.rs
```diff
@@ -6,6 +6,11 @@ use camino::{Utf8Path, Utf8PathBuf};
 use crate::{discover::discover_project, project::utils::FsIoError};
 
 pub fn do_root<P: AsRef<Utf8Path>>(path: P) -> Result<Option<Utf8PathBuf>, Box<FsIoError>> {
-    let root = discover_project(path)?.map(|e| e.root_path().to_owned());
-    Ok(root)
+    discover_project(path)?
+        .map(|e| {
+            let root = e.root_path();
+            root.canonicalize_utf8()
+                .map_err(|err| Box::new(FsIoError::Canonicalize(root.to_owned(), err)))
```
andrius-puksta-sensmetry:
```suggestion
            wrapfs::canonicalize(e.root_path())
```

andrius-puksta-sensmetry (commented):
Implementation is fine, but I really don't like the design and defaults. I'd like to make invalid states unrepresentable, my previous [comment](https://github.com/sensmetry/sysand/pull/411#discussion_r3425732868) details what that would look like. 

Bindings can be more or less freely altered (very few users), so I would like bindings to have my proposed design, while CLI will wait until 0.2.0 to match. Alternatively, we could release this now, but I would still like to make the change on the next breaking release, and breaking the functionality soon after releasing it is not appealing.

bindings/py/src/lib.rs
andrius-puksta-sensmetry: (but python side does, so it's fine)

core/src/commands/sources.rs
```diff
@@ -119,9 +165,67 @@ pub fn find_project_dependencies<Env: ReadEnvironment + Debug + 'static>(
...
+    let provided_iris = match dependencies {
+        Dependencies::Deps => known_std_libs(),
+        _ => HashMap::default(),
+    };
+
+    let resolved = solve_dependencies(requested, env, &provided_iris)?;
+
+    let std_iris: HashSet<fluent_uri::Iri<String>> = known_std_libs()
+        .into_keys()
+        .map(|iri| fluent_uri::Iri::parse(iri).expect("standard library IRIs are valid"))
```
andrius-puksta-sensmetry: `.expect()` takes an error message and prints it verbatim, so this results in
```text
thread '...' panicked at ... : standard library IRIs are valid
```
which is misleading. This is better:
```suggestion
        .map(|iri| fluent_uri::Iri::parse(iri).expect("BUG: invalid std lib IRI"))
```

core/src/commands/sources.rs
```diff
@@ -119,9 +165,67 @@ pub fn find_project_dependencies<Env: ReadEnvironment + Debug + 'static>(
...
+        .collect();
+
+    Ok(resolved
+        .into_iter()
+        .filter(|(iri, _)| match dependencies {
+            Dependencies::None => false,
+            Dependencies::Std => std_iris.contains(iri),
+            Dependencies::Deps | Dependencies::DepsStd => true,
```
andrius-puksta-sensmetry: These can't both be correct. Only `DepsStd` includes everything.

andrius-puksta-sensmetry (approved):

PR #412, opened 2026-06-18
refactor: turn `InterchangeProjectUsageG` into an enum

andrius-puksta-sensmetry:
No behavior changes. This is a preparation for adding more dependency types, which will be variants of the enum. Current plan is something like this:

```rust
pub enum InterchangeProjectUsageG<Iri, VersionReq, Path> {
    /// Legacy, kept for compatibility only
    Resource {
        resource: Iri,
        version_constraint: Option<VersionReq>,
    },
    Url {
        url: Iri,
        publisher: String,
        name: String,
    },
    Path {
        path: Path,
        publisher: String,
        name: String,
    },
    Git {
        git: Iri,
        id: GitId,
        publisher: String,
        name: String,
    },
    Index {
        publisher: String,
        name: String,
        version_constraint: VersionReq,
    },
}

pub enum GitId {
    Rev(String),
    Tag(String),
    Branch(String),
}
```

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -2,6 +2,7 @@
 //
 // config ref: https://docs.renovatebot.com/configuration-options/
 //
+// validate: renovate-config-validator .github/renovate.json5 --strict --no-global
 {
   $schema: "https://docs.renovatebot.com/renovate-schema.json",
   extends: [
@@ -11,6 +12,7 @@
     "helpers:pinGitHubActionDigestsToSemver",
   ],
   minimumReleaseAge: "7 days",
+  abandonmentThreshold: "3 years",
   packageRules: [
     // Group bumps of all non-major dependencies by using two rules, one for
     // situations like 1.2.3, and one for situations like 0.1.2.
--- a/bindings/java/java/src/main/java/com/sensmetry/sysand/model/InterchangeProjectUsage.java
+++ b/bindings/java/java/src/main/java/com/sensmetry/sysand/model/InterchangeProjectUsage.java
@@ -3,29 +3,7 @@
 
 package com.sensmetry.sysand.model;
 
-public class InterchangeProjectUsage {
-
-    private String resource;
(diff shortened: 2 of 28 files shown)


.github/renovate.json5
```diff
@@ -11,6 +12,7 @@
     "helpers:pinGitHubActionDigestsToSemver",
   ],
   minimumReleaseAge: "7 days",
+  abandonmentThreshold: "3 years",
```
andrius-puksta-sensmetry, line 15: Some projects have very infrequent releases, such as `sha2` crate. This prevents "abandonment" warnings from showing up in https://github.com/sensmetry/sysand/issues/48

bindings/java/java/src/main/java/com/sensmetry/sysand/model/InterchangeProjectUsage.java
```diff
@@ -3,29 +3,7 @@
...
-    }
-
-    public void setVersionConstraint(String versionConstraint) {
-        this.versionConstraint = versionConstraint;
-    }
-}
+// TODO(Java 17+): use sealed classes and interfaces 
+// This is meant to be a sealed interface and Sysand only supports
+// usage classes from this package.
+public interface InterchangeProjectUsage {}
```
andrius-puksta-sensmetry, line 9: Note: the conventional way to model a Rust-style enum (i.e. sum type) in Java is via a sealed interface that only allows known variants to implement it via `permits`.

consideRatio (approved):
To not be blocked when making smaller scoped PRs towards a larger is essential in order not loose momentum - im off today but this sounds reasonable.

Another sentence or two about why it helps towards it in the PR description could be valuable.

consideRatio:
I appreciate making PRs smaller scoped and its great, but a failuremode is when review is slow and your work builds on sevaral unmerged PRs which then gets change suggestions etc - so its essential to flush these fast i think.

PR #414, opened 2026-06-19
feat: use consistent path canonicalization everywhere, most relevant on Windows

andrius-puksta-sensmetry:
We use `dunce::canonicalize()` in `wrapfs`, so make all places use it for consistency and better user experience. On Windows, regular `canonicalize()` returns UNC paths (they begin with `\\?\`, which is very unfamiliar to users), while `wrapfs::canonicalize()` strips the UNC prefix where possible.


consideRatio (approved):
Sounds sensible, lets go!

PR #418, opened 2026-06-22
maint: consistently format errors

### LINKED ISSUE #415: Debug sysand package with error in `sysand sources`
There is a failure in https://sysand.com/projects/hugoormo/fibo2sysmlv2/0.1.1/validation/ that I don't understand, indicating that there is something we can improve. The error message should at least be clearer if its not a bug in the client, as its not managing to help for example me to realize whats wrong and what needs to be fixed.

```
RuntimeError: Command ['sysand', 'sources'] failed with exit status 1;
stderr:
warning: SysML v2/KerML standard library packages are omitted by default. If you want to include them, pass `--include-std` flag
error: failed to parse `` as IRI: unexpected character at index 0
note: pass `-v`/`--verbose` to output additional logs
```

andrius-puksta-sensmetry:
We're using `#[source]`/`#[from]` attributes more and more to derive `Error` trait, which means that error sources are sometimes no longer included in `Display`/`{}`/`.to_string()` implementations. Use our own `format_err()`, which prints the whole error sources chain everywhere where we use `.to_string()`/`{}` on an error.

Also provide attribution to a project when its validation fails.
Fixes #415.

### DIFF
--- a/bindings/java/src/lib.rs
+++ b/bindings/java/src/lib.rs
@@ -22,6 +22,7 @@ use sysand_core::{
         utils::wrapfs,
     },
     resolve::{net_utils::create_reqwest_client, standard::standard_resolver},
+    utils::format_err,
     workspace::Workspace,
 };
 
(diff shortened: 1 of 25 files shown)


consideRatio (approved):
Human look through: seems like a good idea and reasonable implemented!

PR #426, opened 2026-07-01
feat: relative directory usage type

andrius-puksta-sensmetry:
This PR adds support for referencing local directory paths as project dependencies, alongside the existing IRI/URL-based usages.

#### New Directory usage type

A new usage variant references a project by a directory path (relative to the declaring project's root) plus its publisher and name. It serialises as `{"dir": ..., "publisher": ..., "name": ...}`. The declared publisher/name must match what's actually found at the path, and there is no version constraint since a directory contains a single version. Only `local_src` projects are supported; KPARs will use another type, to keep resolution unambiguous.

Unlike legacy `resource` usages, where the IRI is both the project's identity and (possibly) its location, typed usages separate the two: identity is always publisher+name, and the source is explicit. This allows avoiding resolution altogether for typed usages. Currently resolution is still done due to large changes required to avoid it; typed usages always resolve to their type.

#### Resolution and solving

Resolvers now receive the full usage (plus an optional base path for relative filesystem paths) instead of a bare IRI. Resolution outcomes distinguish "not found" from "unresolvable", letting a file-resolver miss fall through to other resolvers.

Every usage derives an identifier (a PURL where possible, an URN fallback otherwise; legacy usages keep their IRI). The solver identifies packages by this identifier rather than by version constraint or source, so a package referenced with different constraints/paths by different dependents appears only once in the solution.

#### `experimental` CLI subcommand

Adds `sysand experimental add --dir <path>` and `sysand experimental remove <publisher> <name>`:
- `add` computes a path relative to the current project, reads publisher/name from the dependency, and adds a directory usage. By default it then locks and syncs (`--no-lock`/`--no-sync` opt out), rolling back the project file if resolution fails. No source overrides — typed usages will support everything overrides do, so they can be added directly instead.
- `remove` removes a usage by publisher and name, which will work for all typed usage types.

These are separate from regular `add`/`remove` to avoid breaking changes, and because the UX for other usage types is still unclear. `experimental` commands will likely replace regular commands on next major release, but some accommodation to minimize breaking changes will be required.

#### Other

- Java and Python bindings are adapted to the new usage model, including the new directory usage.

#### Breaking changes

- In the Java and Python bindings, `info` no longer takes a `relative_file_root` argument (it had no effect previously). Java code matching on usage types must also handle the new directory usage class.
- Lockfile interpretation is a bit stricter, but lockfiles generated by sysand remain valid (previously some fields could be missing and the lockfile would still be usable, now it's not).

### DIFF
--- a/core/src/lock.rs
+++ b/core/src/lock.rs
@@ -21,9 +21,10 @@ use typed_path::Utf8UnixPathBuf;
 use crate::{
     config::OverrideSource,
     env::ReadEnvironment,
+    model::InterchangeProjectUsage,
     project::{
         ProjectChecksum,
-        utils::{deserialize_unix_path, serialize_unix_path},
+        utils::{Identifier, deserialize_unix_path, serialize_unix_path},
     },
     utils::{RelativePathKind, RelativeUnixPathError, multiline_array, parse_relative_unix_path},
 };
@@ -240,7 +241,7 @@ impl Lock {
                         return Err(ValidationError::ProjectWithoutId(project.name.clone()));
                     }
                 };
-                return Err(ValidationError::ProjectIdCollision(id.to_string()));
+                return Err(ValidationError::ProjectIdCollision(id.to_owned()));
             }
             for name in &project.exports {
                 if !seen_exports.insert(name) {
@@ -252,7 +253,7 @@ impl Lock {
     }
 
     fn validate_usages(&self) -> Result<(), ValidationError> {
-        let mut iri_versions = HashSet::new();
+        let mut identifier_versions = HashSet::new();
         for project in &self.projects {
             let _ = Version::parse(&project.version).inspect_err(|err| {
                 log::warn!(
@@ -264,13 +265,13 @@ impl Lock {
                     err
                 );
             });
-            for iri in &project.identifiers {
-                iri_versions.insert(iri.clone());
+            for id in &project.identifiers {
+                identifier_versions.insert(id.clone());
             }
         }
         for project in &self.projects {
             for usage in &project.usages {
-                if !iri_versions.contains(usage.inner()) {
+                if !identifier_versions.contains(usage.inner()) {
                     return Err(ValidationError::UnsatisfiedUsage {
                         usage: usage.inner().to_owned(),
                         name: project.name.clone(),
@@ -401,6 +402,48 @@ pub const PROJECT_ENTRIES: &[&str] = &[
     "usages",
     "sources",
 ];
+
+// TODO: decide whether to change lockfile project identifier.
+// Ideally, we would require publisher to be present, then
+// `identifiers` could be removed entirely, and
+// projects would refer to each other by publisher and name only:
+// - usages in .project.json
+// - usages in lockfile
+// - standard libraries (can hardcode them to be treated as
+//   having a publisher `OMG` or similar)
+// Not having to deal with identifiers would be nice,
+// but would require an extremely large refactoring
+//
+// For:
+// - project identity becomes clearly defined for all projects,
+//   useful for deduplication, especially during locking
+// - clear how to communicate to the user about a project (errors, logs)
+// - less stuff in lockfile
+// - less identity information to keep track of
+// - easy dependency replacements (publisher+name is much simpler for the
+//   user than IRIs with their normalization etc.)
+//
+// Against:
+// - can't drop support for projects without a publisher, but
+//   publisher becomes load-bearing and necessary for correctness.
+//   TBD how to resolve this, maybe use a placeholder for all such
+//   projects, like `<unspecified-publisher>`.
+// - requires special treatment for std libs (not a big deal, they're
+//   already specially treated)
+// - populating a lockfile will always require pulling the project
+//   info for index projects, since we need non-normalized publisher
+//   and name for identification (or maybe it would be fine to use
+//   normalized variants in lockfile? they still allow checking
+//   if deps are satisfied. Alternatively, add them to versions.json,
+//   it's backwards compatible)
+//
+// Alternatively, use my previous design here that has a separate
+// `Identifier` type, which is constructed:
+// - `Resource`: just the IRI
+// - `Directory/Kpar`: URL-encoded publisher+name
+// - `Index`: pkg:sysand where possible, else URL-encoded publisher+name
+// - `Git`: TBD (maybe URL, maybe URL-encoded publisher+name)
+// - `Url`: TBD (maybe URL, maybe URL-encoded publisher+name)
 
 /// Fields that might not be set for every project are `Option`
 #[derive(Clone, Eq, Debug, Deserialize, PartialEq)]
@@ -690,25 +733,30 @@ impl Deref for Usage {
     }
 }
 
-impl From<String> for Usage {
-    fn from(resource: String) -> Self {
-        Self(resource)
+impl From<&InterchangeProjectUsage> for Usage {
+    fn from(value: &InterchangeProjectUsage) -> Self {
+        Self(Identifier::from(value).into_string())
     }
 }
 
-impl From<Iri<String>> for Usage {
-    fn from(resource: Iri<String>) -> Self {
-        Self(resource.into_string())
+impl From<InterchangeProjectUsage> for Usage {
+    fn from(value: InterchangeProjectUsage) -> Self {
+        Self(Identifier::from(value).into_string())
     }
 }
 
 impl Usage {
-    pub fn inner(&self) -> &String {
+    pub fn inner(&self) -> &str {
         &self.0
     }
 
-    pub fn into_string(self) -> String {
+    pub fn into_inner(self) -> String {
         self.0
+    }
+
+    /// This should not be used where possible
+    pub fn from_str_unchecked(u: &str) -> Usage {
+        Self(u.to_owned())
     }
 
     fn to_toml(&self) -> Value {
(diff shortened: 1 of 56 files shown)


vytautas-astrauskas-sensmetry:
> It serialises as {"dir": "relative/path", "publisher": "...", "name": "..."}.

Why are `publisher` and `name` fields needed? [In Cargo](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-path-dependencies), as far as I understand, only `path` is used to specify the dependency and name is provided to be able to rename the dependency from the user's side.

andrius-puksta-sensmetry:
Publisher and name are needed for a few reasons:
- consistent UX for different usage types: some usage types (git, index) require publisher+name for obtaining the correct project; others (src path, kpar path, URL) in principle don't, but the UX will be inconsistent if publisher/name are unknown
- not requiring publisher and name for all typed usages will have bad UX for:
  - dependency overrides
  - manual inspection of usage list (also info output): `../../some-dir/aaa/tmp/something` is less informative than `publisher = <pub>, name = <name>, dir = <dir>`
  - error reporting, whenever it relies on usage specification
  - other tools, like IDEs: e.g. if imports are of the form `publisher::name` or something, not having publisher/name explicitly specified in `.project.json` would result in confusing (and unactionable) errors about unknown names if the usage is renamed/moved
- consistency with cargo: cargo [_does_ require name to match actual name](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#renaming-dependencies-in-cargotoml); if a user wants to rename the dep, `package = "original-name"` must be specified. Note that Cargo has good reasons to do it this way; not all of them apply to us:
  - each `use` can be checked from just Cargo.toml (any identifier is fine for this, so the real name is not needed)
  - explicit intent to have that dependency: allows ensuring that the expected project is at the destination if renames/moves happen
  - simpler resolution implementation by having all the info upfront
  - better error reporting if duplicates of the same project from different sources are resolved 
- simpler implementation, especially for error reporting
- push users to have `publisher` set to have typed usages
- simpler project sources deduplication/merging

sysand/tests/cli_exp_add_remove.rs
```diff
@@ -0,0 +1,299 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+// SPDX-FileCopyrightText: © 2025 Sysand contributors <opensource@sensmetry.com>
```
vytautas-astrauskas-sensmetry:
```suggestion
// SPDX-FileCopyrightText: © 2025-2026 Sysand contributors <opensource@sensmetry.com>
```

andrius-puksta-sensmetry: This one should be just 2026, it's a new file.

Regarding the other files, there is no broad consensus on whether to keep the dates updated or just use the year the file was created, so I don't update them.

sysand/tests/cli_exp_add_remove.rs
```diff
@@ -0,0 +1,299 @@
...
+    out.assert().success().stderr(predicate::str::contains(
+        "Adding usage: `untitled/my-dep` from `dep`",
+    ));
+
+    let info_json = std::fs::read_to_string(cwd.join(".project.json"))?;
+    assert_eq!(
+        info_json,
+        r#"{
+  "name": "exp_add_and_remove",
+  "publisher": "untitled",
```
vytautas-astrauskas-sensmetry: What are the reasons to not force the user to specify the `publisher`? Is it friction for the user? Could we reduce it by, for example, having a global config that stores the default value for `publisher`? I am thinking to the experience similar to what git does when the author is not set.

andrius-puksta-sensmetry: Basically to be less annoying. Working with projects without publishing to an index does not really care about the publisher - it will be recorded in project files, but user ~never has to care about it (even new usage types in CLI can mostly be manipulated without knowing the publisher). Also, since publisher is tied to publishing to an index, user does not necessarily know what to choose when creating the project.

Will think about having global config default for this.

sysand/tests/cli_lock.rs
```diff
@@ -687,6 +687,204 @@ fn lock_rejects_non_normalized_sysand_purl() -> Result<(), Box<dyn std::error::E
...
+        let [Source::LocalSrc { src_path, .. }] = project.sources.as_slice() else {
+            panic!("expected a single local source for `{name}`, got: {project:#?}");
+        };
+        assert_eq!(src_path.as_str(), expected_src_path);
+    }
+
+    Ok(())
+}
+
+/// A dependency resolved *from the environment* can declare a directory
```
vytautas-astrauskas-sensmetry, line 726: What does it mean to install a project with file dependencies into an environment? Are the projects referenced by dir usages copied over into the environment, or are they just linked? I would expect that most users for dir usages would expect that they are editable (like in Cargo), which requires them to be linked instead of copied over.

core/src/lock.rs
```diff
@@ -402,9 +403,53 @@ pub const PROJECT_ENTRIES: &[&str] = &[
     "sources",
 ];
 
+// TODO: decide whether to change lockfile project identifier.
+// Ideally, we would require publisher to be present, then
+// `identifiers` could be removed entirely, and
```
vytautas-astrauskas-sensmetry, line 408: Making this change would be great, but I think that it is a bit too early to do it now.

sysand/tests/cli_lock.rs
andrius-puksta-sensmetry: Everything is copied by default, like now. The only projects not copied are "editable" projects (as listed in `env.toml`), which currently are projects from current workspace and those which were `add`ed with `--as-editable`.

andrius-puksta-sensmetry: The distinction between "local" and "editable" is the same as in the lockfile. `local_src` (lockfile equivalent of `Directory` usage)  dependencies by default are treated as non-editable, so lockfile includes their hashes, and so they get installed into the environment. This can be changed, but then we'd probably need to treat all `local_src` deps as editable also in the lockfile.

sysand/tests/cli_exp_add_remove.rs
consideRatio: I think current state or requiring it directly is fine, but a global config seems like a messy UX/DX if implemented.

I'm reasoning that a project aspect like a publisher name belongs in the project, not in a global config, so the only thing a global config then could control would be what sysand init does - but that isnt used often enough to motivate a global config.

If publisher becomes more relevant over time, the more relevant it may be to have publisher be set explicitly early on.

core/src/resolve/combined_tests.rs
```diff
@@ -281,6 +310,70 @@ fn skip_non_semantic_versions() {
     assert_eq!(info.name, "a");
 }
 
+/// A dependency obtained from an environment (or an index/kpar copy) may
+/// declare a `Directory` usage whose relative path only exists in its
```
vytautas-astrauskas-sensmetry, line 314: Seeing this and other edge cases, I am kind of starting to wonder whether we should just require that all transitive dependencies are registry only.

vytautas-astrauskas-sensmetry (dismissed):

sysand/tests/cli_exp_add_remove.rs
andrius-puksta-sensmetry: For now will simply require `--publisher` for `init`. Will consider improvements later.

andrius-puksta-sensmetry: Opened #470.

sysand/tests/cli_lock.rs
andrius-puksta-sensmetry: After a discussion, we decided to keep the current default of treating projects as non-editable by default and thus copying the into the env.

core/src/resolve/combined_tests.rs
andrius-puksta-sensmetry: Requiring transitive deps to be only from registry would be too strict in my opinion, and would be very annoying for local development of a few related projects.

Everything works fine (AFAIK) as long as the environment is not involved. Dependencies of projects in the environment would be cleanly resolvable if env.toml records the original source of the project (so that relative path usages can be resolved correctly), which it currently doesn't. Will make a PR that adds this info to env.toml. Also, env.toml does not record which versions of dependencies of the installed projects (just their identifiers, so it would break with two versions of the same PURL present), this also should be fixed eventually...

vytautas-astrauskas-sensmetry: How much implementation effort each choice would require? If we start with a more restrictive policy now, we can always relax it.

> would be very annoying for local development of a few related projects

This issue should be addressed as soon as we have workspaces.

andrius-puksta-sensmetry: > How much implementation effort each choice would require? If we start with a more restrictive policy now, we can always relax it.

Difficult to say. Dependency source resolution is such an unspecified mess now that achieving any order requires rewiring (at least) the whole solver. It needs to know whether a package is a direct or indirect dependency of root project (to correctly prioritize direct root dependencies), and env resolver (i.e. caching) would still have to be integrated. The only difference is whether to outright reject indirect non-index deps, or try reconciling them with root deps.

Without doing a substantial portion of the work I can't tell how much either option would take :(

vytautas-astrauskas-sensmetry: I see. Then, I leave it for you to decide.

andrius-puksta-sensmetry: Tracked by #439.

vytautas-astrauskas-sensmetry (approved):

PR #440, opened 2026-07-09
feat: support URL templates for index locations ({path} / {path_raw})

### LINKED ISSUE #438: Templated index URLs
I worked on private index repos, and couldnt get a nice story for using gitlab to host a private index without using "classic PAT" tokens that require read access to all repos (bad), or repo admin provided project token shared for non-admin people.

I think though the new beta fine grained PAT could do the trick by granting access to the Files API.

However. Then we must add support for something we dont yet have, the ability to accept a URL structure configured in sysand-index-config.json that allows us to construct an URL like below, where we provide a template to make it work. Crates.io does this as well, while we now just append to the configured index_root seen in the sysand-index-config.json

GET /api/v4/projects/:id/repository/files/:file_path/raw?ref=main

consideRatio:
The feature of allowing an index URL to be a template with `{path}` or `{path_raw}` expanding placeholders, allows use of a  index as git content in a GitLab repo branch.

This feature makes the setup of a GitLab based index align with a GitHub based index with regards to:
- Won't need GitLab/GitHub pages to serve content (GitLab Files API URLs can be used, like GitHub's raw.githubusercontent.com URLs), reducing requirements and complexities
- Won't need GitLab classic PAT / GitLab project scoped access tokens, which both have significant downsides. A classic PAT has read access to all repos, and a project scoped access token can only be created by the index project admins rather than the typical low-level-project-reader that an index reader is expected to be.

With this PR landed, the GitLab private index example repo at https://gitlab.com/sensmetry/public/sysand-private-index can provide simpler and safer simpler instructions for the person setting it up and the people using it.

- fixes #438
- associated client docs PR in https://gitlab.com/sensmetry/internal2/tech/syside/sysand/index-website/-/merge_requests/34

## Vacation discalimer

I'm on vacation, but this was fun work and I lack another constructive hobby at the moment. This PR will however introduce work effort in review etc, so feel free to not spend time on it.

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -61,6 +61,7 @@ walkdir = "2.5.0"
 wasm-bindgen = { version = "0.2.114", default-features = false, optional = true }
 zip = { version = "8.0.0", default-features = false, optional = true, features = ["deflate"] }
 url = { version = "2.5.8", default-features = false }
+percent-encoding = { version = "2.3.2", default-features = false, features = ["alloc"] }
 gix = { version = "0.83.0", default-features = false, optional = true, features = ["blocking-http-transport-reqwest", "blocking-network-client", "worktree-mutation", "sha1"] }
 logos = "0.16.1"
 futures = { version = "0.3.32", default-features = false, features = ["alloc", "async-await"] }
--- a/core/src/auth_tests.rs
+++ b/core/src/auth_tests.rs
@@ -59,3 +59,44 @@ fn basic_globmap_lookup() -> Result<(), Box<dyn std::error::Error>> {
 
     Ok(())
 }
+
+#[test]
+fn globmap_matches_template_expanded_urls() -> Result<(), Box<dyn std::error::Error>> {
+    // The credential glob a user configures for a templated index
+    // (`SYSAND_CRED_X=https://gitlab.com/api/v4/projects/123/**`) must
+    // match the expanded request URLs, where the file path sits mid-URL
+    // percent-encoded and a query string follows.
+    // This glob only fits the *expanded* request URL: it matches on the
+    // percent-encoded file path (`admin%2F...`), which does not exist in
+    // the raw template (where the placeholder still reads `{path}`).
+    let mut builder = GlobMapBuilder::new();
+    builder.add(
+        "https://gitlab.com/api/v4/projects/123/repository/files/admin%2F*/raw?ref=main",
+        1,
+    );
+    let mut globmap = builder.build()?;
+
+    let template = crate::index_location::IndexLocation::parse(
+        "https://gitlab.com/api/v4/projects/123/repository/files/{path}/raw?ref=main",
+    )?;
+    let expanded = template.resolve("admin/proj0/versions.json".split('/'));
+    assert_eq!(
+        expanded.as_str(),
+        "https://gitlab.com/api/v4/projects/123/repository/files/\
+         admin%2Fproj0%2Fversions.json/raw?ref=main"
+    );
+
+    // The raw template does not match the glob...
+    assert!(matches!(
+        globmap.lookup_mut(&template.to_string()),
+        GlobMapResultMut::NotFound
+    ));
+    // ...but the expanded request URL does.
+    if let GlobMapResultMut::Found(_, val) = globmap.lookup_mut(expanded.as_str()) {
+        assert_eq!(*val, 1);
+    } else {
+        panic!("expected credential glob to match expanded template URL");
+    }
+
+    Ok(())
+}
--- a/core/src/commands/publish.rs
+++ b/core/src/commands/publish.rs
@@ -21,6 +21,7 @@ use crate::{
     auth::{ForceBearerAuth, GlobMap, GlobMapResult, HTTPAuthentication},
     env::discovery::{HttpBaseUrlShapeError, validate_http_base_url_shape},
     include::{IncludeError, extract_symbols},
+    index_location::IndexLocation,
     model::{
         InterchangeProjectUsageRaw, InterchangeProjectValidationError, KERML_METAMODEL_PREFIX,
         KerMlChecksumAlg, SYSML_METAMODEL_PREFIX,
@@ -204,7 +205,7 @@ enum SelectedTrustedPublishingProvider<'a> {
 
 pub fn do_publish(
     prepared: PublishPreparation,
-    discovery_root: Url,
+    discovery_root: IndexLocation,
     api_root: Url,
     auth: ForceBearerAuth,
     client: reqwest_middleware::ClientWithMiddleware,
@@ -215,7 +216,7 @@ pub fn do_publish(
     // `api_root`. `discovery_root` is the user-facing URL (what was
     // passed as `--index`) and is kept only so log messages match what
     // the user configured — the actual upload targets `api_root`.
-    let upload_url = build_upload_url(&api_root)?;
+    let upload_url = build_upload_url(&api_root);
     let PublishPreparation {
         norm_publisher: publisher,
         norm_name: name,
@@ -270,30 +271,15 @@ pub fn do_publish(
     map_publish_response(status, &body_bytes, &upload_url_for_log, &response_url)
 }
 
-/// Which root is being validated — selects the error variant so the
-/// message names the spec concept the URL came from.
-#[derive(Debug, Clone, Copy)]
-pub enum EndpointKind {
-    /// User-supplied URL (`--index`).
-    DiscoveryRoot,
-    /// Resolved URL coming back from the discovery document.
-    ApiRoot,
-}
-
-/// Validate the shape of an index-server endpoint URL before the network
-/// step. Applies to both the user-supplied discovery root (pre-discovery)
-/// and the resolved `api_root` that comes back from discovery.
-pub fn validate_endpoint_url_shape(url: &Url, kind: EndpointKind) -> Result<(), PublishError> {
+/// Validate the shape of the resolved `api_root` that comes back from
+/// discovery before the upload step. The user-supplied discovery root is an
+/// [`IndexLocation`], which already enforces the HTTP(S)/no-userinfo
+/// invariant when parsed.
+pub fn validate_api_root_url_shape(url: &Url) -> Result<(), PublishError> {
     let err = |reason: String| -> PublishError {
-        match kind {
-            EndpointKind::DiscoveryRoot => PublishError::InvalidDiscoveryRoot {
-                url: url.as_str().into(),
-                reason,
-            },
-            EndpointKind::ApiRoot => PublishError::InvalidApiRoot {
-                url: url.as_str().into(),
-                reason,
-            },
+        PublishError::InvalidApiRoot {
+            url: url.as_str().into(),
+            reason,
         }
     };
     validate_http_base_url_shape(url).map_err(|e| {
@@ -323,17 +309,14 @@ pub fn validate_endpoint_url_shape(url: &Url, kind: EndpointKind) -> Result<(), 
 }
 
 /// Build the `POST` URL for the publish endpoint from a resolved
-/// `api_root`. The caller is responsible for having resolved the API
-/// root via `sysand-index-config.json`; this function appends the
-/// publish endpoint path to `api_root` as given and does not prepend
-/// any `/api/` segment — that belongs to the API root itself.
-pub fn build_upload_url(api_root: &Url) -> Result<Url, PublishError> {
-    // The `v1/upload` suffix rejection is part of shape validation.
-    validate_endpoint_url_shape(api_root, EndpointKind::ApiRoot)?;
-
-    Ok(crate::env::discovery::with_trailing_slash(api_root.clone())
-        .join(UPLOAD_ENDPOINT_PATH)
-        .unwrap())
+/// `api_root`. Precondition: `api_root` has already been checked with
+/// [`validate_api_root_url_shape`] (done once per publish) and normalized
+/// to end with `/` — both hold for a resolved API root from discovery — so
+/// it is an HTTP(S) base and the endpoint path appends rather than replaces
+/// its last segment. Does not prepend any `/api/` segment; that belongs to
+/// the API root itself.
+pub fn build_upload_url(api_root: &Url) -> Url {
+    api_root.join(UPLOAD_ENDPOINT_PATH).unwrap()
 }
 
 /// Resolve the bearer token used for publishing.
@@ -350,7 +333,7 @@ pub fn resolve_publish_bearer(
     client: &reqwest_middleware::ClientWithMiddleware,
     runtime: &Arc<tokio::runtime::Runtime>,
 ) -> Result<ForceBearerAuth, PublishError> {
-    let upload_url = build_upload_url(api_root)?;
+    let upload_url = build_upload_url(api_root);
     match mode {
         TrustedPublishingMode::Auto => {
             if let Some(provider) = env.trusted_publishing_provider()? {
@@ -454,9 +437,9 @@ fn exchange_oidc_token_for_index_token(
     client: &reqwest_middleware::ClientWithMiddleware,
     runtime: &Arc<tokio::runtime::Runtime>,
 ) -> Result<String, PublishError> {
-    let exchange_url = crate::env::discovery::with_trailing_slash(api_root.clone())
-        .join(TRUSTED_PUBLISHING_EXCHANGE_PATH)
-        .unwrap();
+    // `api_root` is a resolved API root from discovery, already normalized
+    // to end with `/`, so the exchange path appends rather than replaces.
+    let exchange_url = api_root.join(TRUSTED_PUBLISHING_EXCHANGE_PATH).unwrap();
     let body = serde_json::json!({ "token": oidc_token }).to_string();
 
     let response = runtime
@@ -736,9 +719,6 @@ pub enum PublishError {
         license: Box<str>,
         err: spdx::error::ParseError,
     },
-
-    #[error("invalid index URL `{url}` for publish: {reason}")]
-    InvalidDiscoveryRoot { url: Box<str>, reason: String },
 
     #[error("invalid api_root URL `{url}` for publish: {reason}")]
     InvalidApiRoot { url: Box<str>, reason: String },
--- a/core/src/env/discovery.rs
+++ b/core/src/env/discovery.rs
@@ -17,6 +17,7 @@ use crate::{
     auth::HTTPAuthentication,
     env::index::{HttpFetchError, IndexEnvironmentError, MissingPolicy, fetch_json},
     index::iri::parse_iri,
+    index_location::{IndexLocation, IndexLocationError, with_trailing_slash},
 };
 
 const INDEX_PATH: &str = "index.json";
@@ -29,52 +30,42 @@ const META_JSON_FILE: &str = ".meta.json";
 /// discovery step.
 #[derive(Debug, Clone)]
 pub struct ResolvedEndpoints {
-    /// Base URL of the sysand index (where `index.json` lives).
-    pub index_root: url::Url,
-    /// Base URL of the sysand index API (where `v1/upload` lives).
-    pub api_root: url::Url,
+    /// Location of the sysand index (where `index.json` lives): a plain
+    /// base URL or a `{path}` URL template.
+    pub index_root: IndexLocation,
+    /// Base URL of the sysand index API (where `v1/upload` lives). `None`
+    /// when the discovery root is a URL template and the discovery
+    /// document does not supply an `api_root` — such an index is
+    /// read-only from this client's point of view.
+    pub api_root: Option<url::Url>,
 }
 
 impl ResolvedEndpoints {
-    /// Build a `ResolvedEndpoints` that routes both index and API traffic
-    /// at the discovery root itself. Used when the discovery document is
-    /// absent (HTTP 404) or present with neither field set.
-    pub fn flat(discovery_root: url::Url) -> Self {
+    /// Build a `ResolvedEndpoints` that routes index traffic (and, for a
+    /// plain-URL discovery root, API traffic) at the discovery root
+    /// itself. Used when the discovery document is absent (HTTP 404).
+    pub fn flat(discovery_root: IndexLocation) -> Self {
+        let api_root = discovery_root.as_root().cloned();
         Self {
-            index_root: discovery_root.clone(),
-            api_root: discovery_root,
+            index_root: discovery_root,
+            api_root,
         }
     }
 
-    // TODO: url joins will not fail for HTTP URLs, therefore, unwrap
-    fn url_join(url: &url::Url, join: &str) -> Result<url::Url, IndexEnvironmentError> {
-        url.join(join)
-            .map_err(|e| IndexEnvironmentError::JoinURL(url.as_str().into(), join.into(), e))
+    fn resolve<'a>(&self, segments: impl IntoIterator<Item = &'a str>) -> url::Url {
+        self.index_root.resolve(segments)
     }
 
-    pub(crate) fn index_url(&self) -> Result<url::Url, IndexEnvironmentError> {
-        Self::url_join(&self.index_root, INDEX_PATH)
+    pub(crate) fn index_url(&self) -> url::Url {
+        self.resolve([INDEX_PATH])
     }
 
-    pub(crate) fn project_url<S: AsRef<str>>(
-        &self,
-        iri: S,
-    ) -> Result<url::Url, IndexEnvironmentError> {
-        let parsed_iri = parse_iri(iri.as_ref())?;
-        let path = parsed_iri.get_path();
-        Self::url_join(&self.index_root, &format!("{path}/"))
-    }
-
-    /// Per-version directory URL ending with a trailing slash, so that
-    /// `Url::join` treats it as a directory when composing leaf URLs
-    /// (`project.kpar`, `.project.json`, `.meta.json`).
-    pub(crate) fn version_dir_url<S: AsRef<str>, T: AsRef<str>>(
-        &self,
-        iri: S,
-        version: T,
-    ) -> Result<url::Url, IndexEnvironmentError> {
-        let base = self.project_url(iri)?;
-        Self::url_join(&base, &format!("{}/", version.as_ref()))
+    /// The project directory's two path segments for `iri`
+    /// (`[<publisher>, <name>]` or `[_iri, <sha256hex>]`). Returned as
+    /// separate segments so each is encoded independently, avoiding a
+    /// join-then-split round-trip.
+    fn project_rel_segments<S: AsRef<str>>(iri: S) -> Result<[String; 2], IndexEnvironmentError> {
+        Ok(parse_iri(iri.as_ref())?.get_path_segments())
     }
 
     pub(crate) fn kpar_url<S: AsRef<str>, T: AsRef<str>>(
@@ -82,7 +73,13 @@ impl ResolvedEndpoints {
         iri: S,
         version: T,
     ) -> Result<url::Url, IndexEnvironmentError> {
-        Self::url_join(&self.version_dir_url(iri, version)?, KPAR_FILE)
+        let project = Self::project_rel_segments(iri)?;
+        Ok(self.resolve(
+            project
+                .iter()
+                .map(String::as_str)
+                .chain([version.as_ref(), KPAR_FILE]),
+        ))
     }
 
     pub(crate) fn project_json_url<S: AsRef<str>, T: AsRef<str>>(
@@ -90,7 +87,13 @@ impl ResolvedEndpoints {
         iri: S,
         version: T,
     ) -> Result<url::Url, IndexEnvironmentError> {
-        Self::url_join(&self.version_dir_url(iri, version)?, PROJECT_JSON_FILE)
+        let project = Self::project_rel_segments(iri)?;
+        Ok(self.resolve(
+            project
+                .iter()
+                .map(String::as_str)
+                .chain([version.as_ref(), PROJECT_JSON_FILE]),
+        ))
     }
 
     pub(crate) fn meta_json_url<S: AsRef<str>, T: AsRef<str>>(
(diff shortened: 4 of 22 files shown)


core/Cargo.toml
```diff
@@ -61,6 +61,9 @@ walkdir = "2.5.0"
 wasm-bindgen = { version = "0.2.114", default-features = false, optional = true }
 zip = { version = "8.0.0", default-features = false, optional = true, features = ["deflate"] }
 url = { version = "2.5.8", default-features = false }
+# Already a transitive dependency of `url` (same project); used to encode
+# (never decode) relative index paths into URL-template placeholders.
+percent-encoding = { version = "2.3.2", default-features = false, features = ["alloc"] }
```
andrius-puksta-sensmetry:
```suggestion
percent-encoding = { version = "2.3.2", default-features = false, features = ["alloc"] }
```

core/src/auth_tests.rs
```diff
@@ -59,3 +59,32 @@ fn basic_globmap_lookup() -> Result<(), Box<dyn std::error::Error>> {
 
     Ok(())
 }
+
+#[test]
+fn globmap_matches_template_expanded_urls() -> Result<(), Box<dyn std::error::Error>> {
```
andrius-puksta-sensmetry, line 64: Add a test (or modify this one) that globmap matches the expanded URL whose non-expanded version does not match.

core/src/index_location.rs
```diff
@@ -0,0 +1,463 @@
...
+    .remove(b'_')
+    .remove(b'~');
+
+#[derive(Error, Debug)]
+pub enum IndexLocationError {
+    #[error("invalid index URL `{url}`")]
+    InvalidUrl {
+        url: Box<str>,
+        #[source]
+        source: url::ParseError,
```
andrius-puksta-sensmetry:
```suggestion
        source: url::ParseError,
```
`#[source]` is automatically applied to a field named `source`

core/src/index_location.rs
```diff
@@ -0,0 +1,463 @@
...
+    },
+    #[error(
+        "index URL `{url}` is relative; \
+         index locations must be absolute HTTP(S) URLs"
+    )]
+    RelativeUrl { url: Box<str> },
+    #[error(
+        "index URL `{url}` contains `%7B`/`%7D` (percent-encoded braces) — \
+         did you mean a `{{path}}` template? Placeholders must be written \
+         literally, e.g. `.../repository/files/{{path}}/raw?ref=main`"
```
andrius-puksta-sensmetry:
```suggestion
        "index URL `{url}` contains `%7B`/`%7D` (percent-encoded braces);\n\
         did you mean a `{{path}}` template? Placeholders must be written\n\
         literally, e.g. `.../repository/files/{{path}}/raw?ref=main`"
```
And in general no em-dashes in errors. Long messages should be broken up into lines (newlines are `\n\`).

core/src/index_location.rs
```diff
@@ -0,0 +1,463 @@
...
+
+impl PathEncoding {
+    fn placeholder(self) -> &'static str {
+        match self {
+            Self::Encoded => PATH_PLACEHOLDER,
+            Self::Raw => PATH_RAW_PLACEHOLDER,
+        }
+    }
+
+    fn encode(self, rel_path: &str) -> String {
```
andrius-puksta-sensmetry:
```suggestion
    /// `rel_path` cannot contain leading, trailing or multiple consecutive `/`
    fn encode(self, rel_path: &str) -> String {
        debug_assert!(!(rel_path.starts_with('/') || rel_path.ends_with('/'));
```

core/src/index_location.rs
```diff
@@ -0,0 +1,463 @@
...
+                }
+                encoded
+            }
+        }
+    }
+}
+
+/// A URL template with exactly one `{path}` or `{path_raw}` placeholder,
+/// validated at construction. Carried as a string because `url::Url`
+/// percent-encodes `{`/`}` on parse, which would corrupt the placeholder.
```
andrius-puksta-sensmetry:
```suggestion
/// percent-encodes `{` or `}` on parse, which would corrupt the placeholder.
```

core/src/index_location.rs
```diff
@@ -0,0 +1,463 @@
...
+        if s.contains('%') {
+            let lower = s.to_ascii_lowercase();
+            if lower.contains("%7bpath%7d") || lower.contains("%7bpath_raw%7d") {
+                return Err(IndexLocationError::PreEncodedPlaceholder { url: s.into() });
+            }
+        }
+        match url::Url::parse(s) {
+            // Normalize the trailing slash once here so `resolve` can
+            // join without cloning. Non-hierarchical URLs (`mailto:`)
+            // cannot be normalized and are rejected later by `resolve`.
```
andrius-puksta-sensmetry: Move at least the `validate_http_base_url_shape` here. Checks should be done as early as possible.

core/src/index_location.rs
```diff
@@ -0,0 +1,463 @@
...
+                if root.cannot_be_a_base() {
+                    return Err(ResolveUrlError::NotABase {
+                        root: root.as_str().into(),
+                    });
+                }
+                root.join(rel_path).map_err(|source| ResolveUrlError::Join {
+                    root: root.as_str().into(),
+                    path: rel_path.into(),
+                    source,
+                })
```
andrius-puksta-sensmetry: All this error handling can be removed once this type enforces http/https URLs.

core/src/index_location.rs
```diff
@@ -0,0 +1,463 @@
...
+    /// spelling, into the placeholder and parse the result.
+    pub fn expand(&self, rel_path: &str) -> Result<url::Url, ResolveUrlError> {
+        let encoded = self.encoding.encode(rel_path);
+        let expanded = self.raw.replacen(self.encoding.placeholder(), &encoded, 1);
+        url::Url::parse(&expanded).map_err(|source| ResolveUrlError::Expand {
+            template: self.raw.as_str().into(),
+            path: rel_path.into(),
+            source,
+        })
+    }
```
andrius-puksta-sensmetry: 1. Return `url::Url` directly here. URL will not fail parsing, since it's valid by construction, and adding a (properly encoded) segment cannot (AFAIK) change URL validity, so `unwrap()` is warranted.
2. To avoid unnecessary checks of the shape of `rel_path` in `encode` (e.g. leading, trailing or multiple slashes), take an `Iterator` over the segments. Then each segment can be encoded directly without worrying about slashes.

core/src/index_location.rs
```diff
@@ -0,0 +1,463 @@
...
+}
+
+/// A URL template with exactly one `{path}` or `{path_raw}` placeholder,
+/// validated at construction. Carried as a string because `url::Url`
+/// percent-encodes `{`/`}` on parse, which would corrupt the placeholder.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct IndexUrlTemplate {
+    raw: String,
+    encoding: PathEncoding,
+}
```
andrius-puksta-sensmetry, line 165:
```suggestion
pub struct IndexUrlTemplate {
    prefix: String,
    suffix: String
    encoding: PathEncoding,
}
```
And in `expand()` do `prefix + (each encoded path segment) + suffix`. This is faster and a bit clearer.

core/src/env/discovery.rs
```diff
@@ -248,48 +275,47 @@ pub async fn fetch_index_config<P: HTTPAuthentication>(
...
-                value,
-                source: url::ParseError::RelativeUrlWithoutBase,
-            }) => Err(DiscoveryError::RelativeUrl { url, field, value }),
-            other => other,
-        };
+    // `index_root` may itself be a URL template; `api_root` may not
+    // (uploads are not file fetches, so templating it is meaningless).
+    let index_root = match raw.index_root {
+        None => discovery_location.clone(),
+        Some(s) if is_template_syntax(&s) => {
```
andrius-puksta-sensmetry: This does not check for percent-encoded `{path}`. Is this intentional?

core/src/env/discovery.rs
```diff
@@ -207,33 +217,50 @@ fn discovery_shape_error(
...
-    })?;
+    // Normalize a plain discovery root so relative-path resolution treats
+    // it as a directory. Validate its shape before issuing any request;
+    // templates were already shape-validated when parsed.
+    let discovery_location = match discovery_root {
+        IndexLocation::Root(root) => {
+            validate_http_base_url_shape(root)
+                .map_err(|error| discovery_shape_error(root, "<discovery_root>", root, error))?;
+            IndexLocation::Root(with_trailing_slash(root.clone()))
+        }
```
andrius-puksta-sensmetry: `IndexLocation` is responsible for enforcing its validity invariants.

core/src/env/discovery.rs
```diff
@@ -248,48 +275,47 @@ pub async fn fetch_index_config<P: HTTPAuthentication>(
...
+    let index_root = match raw.index_root {
+        None => discovery_location.clone(),
+        Some(s) if is_template_syntax(&s) => {
+            IndexLocation::parse(&s).map_err(|source| DiscoveryError::InvalidLocation {
+                url: config_url.as_str().into(),
+                field: "index_root",
+                source,
+            })?
+        }
+        Some(s) => IndexLocation::Root(parse_base_url("index_root", s)?),
```
andrius-puksta-sensmetry: I'd say the `Some()` case should directly call `IndexLocation::Parse`. Since the validity invariants are the same for discovery root and index root, they should be enforced in the same place.

sysand/src/commands/publish.rs
```diff
@@ -36,7 +36,10 @@ pub fn command_publish(
     // Reject obviously-malformed discovery-root URLs (bad scheme,
     // query/fragment components) before issuing any network request —
     // a config typo should not cost a DNS lookup + connect attempt.
-    validate_endpoint_url_shape(&index, EndpointKind::DiscoveryRoot)?;
+    // Template locations were fully validated at argument parse time.
+    if let Some(root) = index.as_root() {
+        validate_endpoint_url_shape(root, EndpointKind::DiscoveryRoot)?;
+    }
```
andrius-puksta-sensmetry: `IndexLocation` is responsible for enforcing its validity invariants.

andrius-puksta-sensmetry (changes requested):
Parsing and validation should happen in a single place for each index-related URL, in this case it should be enforced by `IndexLocation`.

Also no point in `format!`ing `rel_path` just to immediately `split` it; use an `Iterator` instead.

consideRatio:
@andrius-puksta-sensmetry thank you for working this! I'm letting AI address the feedback using this prompt, also considering past PR feedback you've given distilled by AI previously:

>I want you to do cycles, and once a cycle is complete, sleep 5 minutes, and check back in for new updates.
>
>Here is the description of a cycle:
>
>1. Use `gh` to read the feedback we've received
>2. Address everything
>3. Commit and push
>4. Provide a to-the-point comment about what you pushed - disclaiming its written by your automation - summary of what you have addressed, linking to each comment, allowing the summary to be collapsable in githubs UI by use of `<details> <summary>description of whats collapsed</summary> collapsed content here </details>`
>5. Disclaim the automation running as well, and enable him to turn it off by writing a comment in a certain way.

consideRatio:
> 🤖 **Automated update** — this comment and the pushed changes were prepared by an automation assistant on Erik's behalf. See the note at the bottom to pause it.

Pushed `bd96685` addressing the latest review round. The through-line: **`IndexLocation` now enforces its own validity invariants at parse time** (absolute HTTP(S), no userinfo — for plain roots as well as templates), which lets downstream shape checks, error handling, and the `ResolveUrlError` type be deleted.

**Highlights**
- `expand()` / `resolve()` are now infallible (valid by construction) and operate over path *segments*, so per-segment encoding needs no slash-shape checks.
- `IndexUrlTemplate` stores `prefix` / `suffix` / `encoding` and expands as `prefix + encoded-segments + suffix`.
- Discovery `index_root` (plain or templated) and the discovery root are both validated through `IndexLocation::parse` — one place — which also rejects percent-encoded `%7Bpath%7D`.
- Publish CLI drops the now-redundant pre-discovery shape check; `EndpointKind` collapses to `validate_api_root_url_shape`.

<details>
<summary>Per-comment mapping (13 threads)</summary>

| Review comment | What changed |
|---|---|
| [source auto-attr](https://github.com/sensmetry/sysand/pull/440#discussion_r3558465535) | Dropped explicit `#[source]` on `source`-named fields |
| [em-dash / line breaks](https://github.com/sensmetry/sysand/pull/440#discussion_r3558474044) | Rewrote `PreEncodedPlaceholder` message with `\n\` line breaks, no em-dash |
| [`IndexUrlTemplate` doc](https://github.com/sensmetry/sysand/pull/440#discussion_r3558835320) | Doc updated for the `prefix`/`suffix` representation |
| [prefix/suffix/encoding struct](https://github.com/sensmetry/sysand/pull/440#discussion_r3568708413) | `IndexUrlTemplate { prefix, suffix, encoding }`; expand = concat |
| [return `url::Url` + segment iterator](https://github.com/sensmetry/sysand/pull/440#discussion_r3568687402) | `expand()` returns `url::Url` (unwrap, valid by construction), takes segments |
| [`encode` slash checks](https://github.com/sensmetry/sysand/pull/440#discussion_r3558828624) | Superseded — segment-based encoding removes the slash-shape concern entirely |
| [move shape check earlier](https://github.com/sensmetry/sysand/pull/440#discussion_r3559025953) | `IndexLocation::parse` enforces HTTP(S)/no-userinfo for roots too |
| [remove error handling](https://github.com/sensmetry/sysand/pull/440#discussion_r3559036369) | `resolve` infallible; `ResolveUrlError` deleted |
| [invariants owned by type (discovery root)](https://github.com/sensmetry/sysand/pull/440#discussion_r3568767425) | Removed inline `validate_http_base_url_shape` in discovery |
| [`Some()` → `IndexLocation::parse`](https://github.com/sensmetry/sysand/pull/440#discussion_r3568781528) | `index_root` routed through `IndexLocation::parse` |
| [percent-encoded `{path}` check](https://github.com/sensmetry/sysand/pull/440#discussion_r3568731261) | Unifying via `parse` also rejects `%7Bpath%7D` in discovery |
| [invariants owned by type (publish)](https://github.com/sensmetry/sysand/pull/440#discussion_r3568790784) | Dropped pre-discovery discovery-root check; `v1/upload` guard stays on `api_root` |
| [globmap discriminating test](https://github.com/sensmetry/sysand/pull/440#discussion_r3557683698) | Test now asserts the glob matches the *expanded* URL but **not** the raw template |

</details>

<details>
<summary>Verification</summary>

- `cargo build --workspace` ✅
- `cargo test -p sysand-core` — 164 passed ✅
- `cargo test -p sysand --test cli_publish` — 29 passed ✅
- `cargo fmt --check` / `cargo clippy` — clean ✅

Two behavior notes worth a look: `--index ftp://…` is now rejected at argument-parse time (clearer, earlier), and a pasted `…/v1/upload` `--index` is now rejected after discovery on the resolved `api_root` rather than pre-discovery — the guard still fires before any upload.

</details>

---
<sub>🤖 This is an automated assistant running in cycles on Erik's machine (read feedback → address → push → summarize). To pause it, post a comment containing the exact text <code>/automation off</code> and it will stop on its next cycle.</sub>

core/src/env/discovery.rs
```diff
@@ -29,84 +30,79 @@ const META_JSON_FILE: &str = ".meta.json";
...
     }
 
     pub(crate) fn kpar_url<S: AsRef<str>, T: AsRef<str>>(
         &self,
         iri: S,
         version: T,
     ) -> Result<url::Url, IndexEnvironmentError> {
-        Self::url_join(&self.version_dir_url(iri, version)?, KPAR_FILE)
+        let project = Self::project_rel_path(iri)?;
+        Ok(self.resolve(project.split('/').chain([version.as_ref(), KPAR_FILE])))
```
andrius-puksta-sensmetry: `project_rel_path` builds a path which then immediately gets `split('/')`. Since it always returns two segments, either return a tuple of strings, or an iterator.

consideRatio:
> 🤖 **Automated update** — prepared by an automation assistant on Erik's behalf. See the note at the bottom to pause it.

Pushed `e5eab16` addressing the follow-up.

- [**`project_rel_path` join-then-split**](https://github.com/sensmetry/sysand/pull/440#discussion_r3570205762) — Added `ParsedIri::get_path_segments() -> [String; 2]` (the two-segment count is now a type-level invariant); `get_path()` delegates to it via `join("/")`, and `ResolvedEndpoints` resolves directly over the segments. No more building `<publisher>/<name>` only to `split('/')` it back at each call site.

<details>
<summary>Verification</summary>

- `cargo build --workspace` ✅
- `cargo test -p sysand-core` — 164 passed ✅ · `cargo test -p sysand --test cli_publish` — 29 passed ✅
- `cargo fmt --check` / `cargo clippy` — clean ✅

</details>

---
<sub>🤖 This is an automated assistant running in cycles on Erik's machine (read feedback → address → push → summarize). To pause it, post a comment containing the exact text <code>/automation off</code> and it will stop on its next cycle.</sub>

core/src/env/discovery.rs
```diff
@@ -207,33 +221,41 @@ fn discovery_shape_error(
...
+    discovery_root: &IndexLocation,
 ) -> Result<ResolvedEndpoints, DiscoveryError> {
-    validate_http_base_url_shape(discovery_root).map_err(|error| {
-        discovery_shape_error(discovery_root, "<discovery_root>", discovery_root, error)
-    })?;
+    // Normalize a plain discovery root so relative-path resolution treats
+    // it as a directory. `IndexLocation` already enforced its shape
+    // invariants (absolute HTTP(S), no userinfo) when parsed.
+    let discovery_location = match discovery_root {
+        IndexLocation::Root(root) => IndexLocation::Root(with_trailing_slash(root.clone())),
```
andrius-puksta-sensmetry: This is the responsibility of `IndexLocation`.

core/src/commands/publish.rs
```diff
@@ -329,9 +315,9 @@ pub fn validate_endpoint_url_shape(url: &Url, kind: EndpointKind) -> Result<(),
 /// any `/api/` segment — that belongs to the API root itself.
 pub fn build_upload_url(api_root: &Url) -> Result<Url, PublishError> {
     // The `v1/upload` suffix rejection is part of shape validation.
-    validate_endpoint_url_shape(api_root, EndpointKind::ApiRoot)?;
+    validate_api_root_url_shape(api_root)?;
 
-    Ok(crate::env::discovery::with_trailing_slash(api_root.clone())
+    Ok(with_trailing_slash(api_root.clone())
```
andrius-puksta-sensmetry: It is the responsibility `ResolvedEndpoints` to ensure that `api_root` is is correct and with a trailing slash.

consideRatio:
> 🤖 **Automated update** — prepared by an automation assistant on Erik's behalf. See the note at the bottom to pause it.

Pushed `50f579b` addressing the follow-up.

- [**discovery-root trailing-slash is `IndexLocation`'s responsibility**](https://github.com/sensmetry/sysand/pull/440#discussion_r3570509632) — Dropped the `with_trailing_slash(root.clone())` re-normalization in `fetch_index_config`. The discovery root is already a parsed `IndexLocation`, whose `Root` variant guarantees the trailing slash (via `IndexLocation::parse`), so it arrives normalized — the function now just clones it. Also updated the one discovery test that constructed `IndexLocation::Root(base)` directly to go through `IndexLocation::parse(...)`, so nothing reaches discovery via a raw enum construction that bypasses the invariant.

<details>
<summary>Verification</summary>

- `cargo build --workspace` ✅
- `cargo test -p sysand-core` — 164 passed ✅ · `cargo test -p sysand --test cli_publish` — 29 passed ✅
- `cargo fmt --check` / `cargo clippy` — clean ✅

</details>

---
<sub>🤖 This is an automated assistant running in cycles on Erik's machine (read Andrius's feedback → address → push → summarize). To pause it, post a comment containing the exact text <code>/automation off</code> and it will stop on its next cycle.</sub>

core/src/commands/publish.rs
```diff
@@ -454,7 +440,7 @@ fn exchange_oidc_token_for_index_token(
     client: &reqwest_middleware::ClientWithMiddleware,
     runtime: &Arc<tokio::runtime::Runtime>,
 ) -> Result<String, PublishError> {
-    let exchange_url = crate::env::discovery::with_trailing_slash(api_root.clone())
+    let exchange_url = with_trailing_slash(api_root.clone())
```
andrius-puksta-sensmetry: Again, this is not the place to do this. In general `with_trailing_slash` usages should be audited, since most of them are called on the same URLs in various places.

core/src/env/discovery.rs
```diff
@@ -29,84 +30,92 @@ const META_JSON_FILE: &str = ".meta.json";
...
-    pub api_root: url::Url,
+    /// Location of the sysand index (where `index.json` lives): a plain
+    /// base URL or a `{path}` URL template.
+    pub index_root: IndexLocation,
+    /// Base URL of the sysand index API (where `v1/upload` lives). `None`
+    /// when the discovery root is a URL template and the discovery
+    /// document does not supply an `api_root` — such an index is
+    /// read-only from this client's point of view.
+    pub api_root: Option<url::Url>,
 }
```
andrius-puksta-sensmetry, line 41: It is the resoponsibility of this struct to ensure that `api_root` is valid (for its purpose) and with a trailing slash (i.e. ensure it passes validate_api_root_url_shape). No other place should perform validation of this. In cases where it is known `api_root` will not be used (and thus there is no need to parse/normalize it), indicate it via the type somehow or just have `None` here. Types should enforce their internal invariants.

consideRatio:
> 🤖 **Automated update** — prepared by an automation assistant on Erik's behalf. See the note at the bottom to pause it.

Pushed `bbf2b79` addressing the follow-up.

- [**audit `with_trailing_slash` usages on `api_root`**](https://github.com/sensmetry/sysand/pull/440#discussion_r3570550769) — `api_root` was being re-normalized with `with_trailing_slash` in both `build_upload_url` and the OIDC token exchange. It's already normalized once at its source (discovery resolves it via `parse_base_url` / the `IndexLocation` root invariant), so both re-normalizations are removed and those functions now `join` directly onto the resolved `api_root`. Documented the "already ends with `/`" precondition on `build_upload_url`.

<details>
<summary>Coverage moved to the source</summary>

The old `build_upload_url` test that passed a slash-less `https://example.org/api` was really exercising normalization — that responsibility now lives in discovery, so it's replaced by a new discovery test (`discovery_normalizes_api_root_trailing_slash`) asserting a discovery-document `api_root` without a trailing slash resolves to one that ends with `/`.

Remaining `with_trailing_slash` call sites after the audit: `IndexLocation::parse` (establishes the `Root` invariant — the canonical spot) and discovery's `parse_base_url` (normalizes `api_root` at resolution). No consumer re-normalizes anymore.

</details>

<details>
<summary>Verification</summary>

- `cargo build --workspace` ✅
- `cargo test -p sysand-core --features networking,filesystem` — 342 passed ✅ (these tests are feature-gated; the default `cargo test -p sysand-core` skips them)
- `cargo test -p sysand --test cli_publish` — 29 passed ✅
- `cargo fmt --check` / `cargo clippy --features networking,filesystem` — clean ✅

</details>

---
<sub>🤖 This is an automated assistant running in cycles on Erik's machine (read Andrius's feedback → address → push → summarize). To pause it, post a comment containing the exact text <code>/automation off</code> and it will stop on its next cycle.</sub>

core/src/index_location.rs
andrius-puksta-sensmetry: Still plenty of places where long messages are not broken up into lines (just a ` \` will not create a newline in the output; `\n\` has to be used).

core/src/index_location.rs
```diff
@@ -0,0 +1,442 @@
...
+    /// `design/index-protocol.md` §5), so they cannot smuggle `..` or URL
+    /// syntax into the result. Infallible, since both variants were
+    /// shape-validated at construction.
+    pub fn resolve<'a>(&self, segments: impl IntoIterator<Item = &'a str>) -> url::Url {
+        match self {
+            Self::Root(root) => {
+                let rel_path = segments.into_iter().collect::<Vec<_>>().join("/");
+                root.join(&rel_path)
+                    .expect("BUG: HTTP(S) base joined with a trusted relative index path")
+            }
```
andrius-puksta-sensmetry, line 335: This is really inefficient. First a string is built, then that same string is parsed in `join()`. Instead of this, clone the URL and append each segment to it.

consideRatio:
> 🤖 **Automated update** — prepared by an automation assistant on Erik's behalf. See the note at the bottom to pause it.

Pushed `bf131d6` addressing the follow-up.

- [**`\n\` for long error messages**](https://github.com/sensmetry/sysand/pull/440#discussion_r3570687207) — Swept the remaining long messages that used a plain ` \` line continuation (which renders as one line) and switched them to `\n\` so they break into actual output lines. Applied consistently across the touched modules: all long `IndexLocationError` variants, the `DiscoveryError` URL-shape variants, and the templated-index publish `bail!`. Breaks are at clause boundaries (usually the `;`).

<details>
<summary>Scope & guardrails</summary>

- Core `PublishError` messages already used `\n\` — left as-is.
- Left short single-clause messages (e.g. `InvalidUrl`, `InvalidLocation`) on one line, and left the pre-existing `without [PATH]` `bail!` alone (not part of this PR's diff).
- Verified no test's `.contains(...)` substring is split by a newline — breaks sit at clause boundaries, asserted substrings stay intact on one line.

</details>

<details>
<summary>Verification</summary>

- `cargo build --workspace` ✅
- `cargo test -p sysand-core --features networking,filesystem` — 342 passed ✅
- `cargo test -p sysand --test cli_publish` — 29 passed ✅
- `cargo fmt --check` / `cargo clippy --features networking,filesystem` — clean ✅

</details>

---
<sub>🤖 This is an automated assistant running in cycles on Erik's machine (read Andrius's feedback → address → push → summarize). To pause it, post a comment containing the exact text <code>/automation off</code> and it will stop on its next cycle.</sub>

core/src/commands/publish.rs
```diff
@@ -323,17 +309,15 @@ pub fn validate_endpoint_url_shape(url: &Url, kind: EndpointKind) -> Result<(),
...
+/// segment — that belongs to the API root itself.
 pub fn build_upload_url(api_root: &Url) -> Result<Url, PublishError> {
     // The `v1/upload` suffix rejection is part of shape validation.
-    validate_endpoint_url_shape(api_root, EndpointKind::ApiRoot)?;
+    validate_api_root_url_shape(api_root)?;
 
-    Ok(crate::env::discovery::with_trailing_slash(api_root.clone())
-        .join(UPLOAD_ENDPOINT_PATH)
-        .unwrap())
+    Ok(api_root.join(UPLOAD_ENDPOINT_PATH).unwrap())
```
andrius-puksta-sensmetry: This function is called twice for a single publish. `validate_api_root_url_shape` should be done once, preferably enforced by `ResolvedEndpoints`.

andrius-puksta-sensmetry: (and so `ResolvedEndpoints` fields should probably be made private to ensure invariants are preserved)

consideRatio:
> 🤖 **Automated update** — prepared by an automation assistant on Erik's behalf. See the note at the bottom to pause it.

Pushed `eeb804b` addressing the follow-up.

- [**validate `api_root` shape once per publish**](https://github.com/sensmetry/sysand/pull/440#discussion_r3570805261) — `build_upload_url` both validated and joined, and it was called from both `resolve_publish_bearer` and `do_publish`, so `validate_api_root_url_shape` ran twice. Split the concerns: `build_upload_url` is now a pure infallible join (precondition — a validated, normalized `api_root` — documented), and `validate_api_root_url_shape` is called exactly once at the publish command boundary, right after `api_root` is resolved.

<details>
<summary>On "enforced by `ResolvedEndpoints`"</summary>

I validated once where publish consumes `ResolvedEndpoints.api_root` rather than inside `ResolvedEndpoints` itself, for two reasons — happy to move it if you'd prefer:

- **Layering:** `ResolvedEndpoints` lives in `env::discovery`; `validate_api_root_url_shape` (and its `PublishError`, including the `v1/upload` guard) lives in `commands::publish`, which depends on `env`, not vice versa. A method on `ResolvedEndpoints` would invert that dependency.
- **Read vs. publish:** `api_root` is only used for publishing; reads never touch it. Enforcing the full publish shape (no query/fragment, not `v1/upload`) at discovery time would make a *read* against an index with an odd `api_root` fail on a publish-only concern.

</details>

<details>
<summary>Tests</summary>

The `build_upload_url_rejects_*` tests became `validate_api_root_rejects_*` (validation moved, so its rejection tests moved with it), keeping coverage of the upload-endpoint, query/fragment, non-hierarchical, and userinfo cases.

- `cargo test -p sysand-core --features networking,filesystem` — 339 passed ✅
- `cargo test -p sysand --test cli_publish` — 29 passed ✅
- `cargo fmt --check` / `cargo clippy --features networking,filesystem` — clean ✅

</details>

---
<sub>🤖 This is an automated assistant running in cycles on Erik's machine (read Andrius's feedback → address → push → summarize). To pause it, post a comment containing the exact text <code>/automation off</code> and it will stop on its next cycle.</sub>

andrius-puksta-sensmetry (approved):
Looking good. When the remaining nit is resolved, this is good to go.

consideRatio:
> 🤖 **Automated update** — prepared by an automation assistant on Erik's behalf. See the note at the bottom to pause it.

Thanks for the approval! Pushed `be46758` resolving the remaining nit.

- [**avoid building-then-reparsing in `resolve`**](https://github.com/sensmetry/sysand/pull/440#discussion_r3570736611) — The `Root` branch of `IndexLocation::resolve` built a `/`-joined `String` and reparsed it in `Url::join`. It now appends each segment directly to a clone of the base via `path_segments_mut().pop_if_empty().extend(segments)` (popping the trailing empty segment first, since the root ends with `/`, to avoid a doubled slash). Same output, no reparse.

<details>
<summary>Note: this nit was briefly missed — sorry</summary>

This comment landed ~20s before one of my earlier summary comments and slipped through a gap in how the automation tracked "already seen" comments (it was keyed off my own last-post time). I've corrected that so reviewer comments in that window aren't skipped again.

</details>

<details>
<summary>Verification</summary>

Exact-URL assertions (`index_location` + discovery tests) confirm the output is unchanged.

- `cargo test -p sysand-core --features networking,filesystem` — 339 passed ✅
- `cargo test -p sysand --test cli_publish` — 29 passed ✅
- `cargo fmt --check` / `cargo clippy --features networking,filesystem` — clean ✅

</details>

---
<sub>🤖 This is an automated assistant running in cycles on Erik's machine (read Andrius's feedback → address → push → summarize). To pause it, post a comment containing the exact text <code>/automation off</code> and it will stop on its next cycle.</sub>

core/src/index_location.rs
consideRatio: 🤖 (automation) Done in be46758 — `resolve` now appends each segment to a clone of the base via `path_segments_mut().pop_if_empty().extend(segments)`, no string-build-then-reparse. Exact-URL test assertions confirm unchanged output.

consideRatio:
Great work on this @andrius-puksta-sensmetry!!! The PR branch is rebased on main with a signed commit and re-tested against a private gitlab index setup with a tightly scoped personal access token and it worked great after the changes.

consideRatio:
Failed to sign the commit correctly, now I managed to do it - tests passed!

PR #449, opened 2026-07-14
fix(publish): surface index error details, drop redundant warn-then-error

consideRatio:
This helps users fix their misconfigurations better, such as not having a protected branch or similar. The trusted publishing requirements on the sysand.com index server is quite stringent, and requires that for example. Instead of presenting just 403 error, this should help the user figure out what to fix.

The removed URL from the removed warning can be found if needed using debug logging still.

---

The trusted-publishing OIDC token exchange discarded the response body
on a non-success status, so users saw only "HTTP status 403" with no
cause. Extract the server-provided `{"error": ...}` message via the
existing error_body_to_string helper and include it in the error.

The upload endpoint already returned the server detail in its 4xx error
variants, but its catch-all branch logged a warning immediately before
returning the error, duplicating what the propagated error already
reports. Remove that warn (and the now-unused url parameters); the
returned error is sufficient on its own.

### DIFF
--- a/core/src/commands/publish.rs
+++ b/core/src/commands/publish.rs
@@ -268,7 +268,7 @@ pub fn do_publish(
         status
     );
 
-    map_publish_response(status, &body_bytes, &upload_url_for_log, &response_url)
+    map_publish_response(status, &body_bytes)
 }
 
 /// Validate the shape of the resolved `api_root` that comes back from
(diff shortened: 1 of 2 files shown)


andrius-puksta-sensmetry (approved):
Thanks!

andrius-puksta-sensmetry:
I'll make a release tomorrow morning.

PR #453, opened 2026-07-19
feat: persist index credentials in the OS keyring (sysand auth)

consideRatio:
I did initial planning of the UX and rough implementation for this feature (in https://github.com/sensmetry/sysand/issues/437#issuecomment-5011192677), then but Andrius did an amazing review effort getting it into a mergable state.

- fixes #437 

## PR description

Implements credential storage and the sysand auth command family
(login, logout, status, whoami), v1 of the design settled in #437 and
maintained in design/credential-storage.md on this branch.

Store an index credential once with sysand auth login and sysand
reuses it across runs on Windows, macOS, and Linux, with the secret
held by the OS keyring, never a plaintext file. Bearer-only v1;
SYSAND_CRED_* environment variables keep working unchanged and take
precedence, so CI overrides an interactive login.

Command surface:

- auth login [INDEX_URL] [--token-stdin]: hidden prompt or stdin,
  never an inline argument; resolves the default-index chain when no
  URL is given and always echoes the target before reading a secret.
  On hosts without a usable keyring backend it refuses to persist
  and prints exact SYSAND_CRED_* lines with a <token> placeholder.
- auth logout [INDEX_URL]: removes a stored login; idempotent.
- auth status: one unified, secret-free view of stored and env
  credentials, with per-surface validation claims, expiry,
  env-shadowing, and a default-index marker.
- auth whoami [INDEX_URL]: live identity check against v1/whoami
  using runtime credential selection (env over stored, naming the
  source used); read-only.

Storage is a single versioned JSON blob in one OS-keyring entry,
fail-closed on unreadable or newer-versioned blobs, guarded by a
cross-process file lock, with platform size limits mapped to clear
errors. The OS backend sits behind a non-default keyring feature;
record types and the LockedBlobStore seam are unconditional core, so
the js/wasm binding stays green.

Consumption is lazy: the existing env policy runs first, and only an
auth-relevant 4xx triggers one cached blob read per process and a
forced bearer retry when a stored glob matches. No matching record
returns the original response with zero extra requests. Local,
offline, public, and never-logged-in flows never touch the keyring.

Login validation is always on and discovery-first: per-surface
probes (read via HEAD, whoami via GET) exercise the credential only
when the unauthenticated baseline was 4xx; 429 and redirects are
never verdicts. Store if any exercised surface accepted; refuse only
when at least one rejected and none accepted; nothing exercised
stores as "stored, not validated". A successful whoami persists
subject, token name/prefix, and expiry for auth status. Basic-auth
[...]

PR #454, opened 2026-07-19
ci: pin the prek job's Rust toolchain to rust-toolchain.toml

consideRatio:
The prek job's `language: rust` hooks do not use the toolchain pinned by `rust-toolchain.toml`: prek picks a toolchain itself from what is installed, preferring the runner image's latest stable. When a newer stable clippy adds an auto-fixable lint, the `cargo clippy --fix` hook rewrites files and fails the gate with "files were modified". This hit #453 after the runner image moved to 1.97, forcing it to carry four unrelated lint fixes.

Two changes, verified locally by reproducing the failure and confirming the fix against prek's trace output:

- `language_version: "1.96"` on the rust hooks in `.pre-commit-config.yaml`: prek's own selector, the part that actually pins.
- A `rustup toolchain install` step in the workflow so the pinned toolchain (with the rustfmt/clippy components) exists for prek to find. Without it, prek downloads a profile-minimal toolchain lacking those components and cargo silently falls through to another toolchain's binaries.

The pin now lives in two places; `rust-toolchain.toml` and the hook config cross-reference each other so they get bumped together.

---

https://claude.ai/code/session_01EfyrkJkeo3mRngVoppJjUN

### DIFF
--- a/.github/workflows/prek.yml
+++ b/.github/workflows/prek.yml
@@ -21,6 +21,14 @@ jobs:
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
       - uses: astral-sh/setup-uv@d31148d669074a8d0a63714ba94f3201e7020bc3 # v8.3.0
+
+      # prek's rust hooks use their language_version pin, not
+      # rust-toolchain.toml; preinstall the pinned toolchain so prek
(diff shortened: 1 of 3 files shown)


andrius-puksta-sensmetry (approved):

andrius-puksta-sensmetry:
(commit has to be signed)

PR #456, opened 2026-07-19
fix(test): stop the load_configs test reading the developer's real config

consideRatio:
`config::local_fs::tests::load_configs` calls `load_configs`, which merges the user configuration from `dirs::config_dir()` before asserting equality against a fixture. The assertion therefore depends on whatever `~/.config/sysand/sysand.toml` happens to contain: green on CI runners (no user config), red on any developer machine where sysand is actually used, with the developer's own index entries showing up in the assertion diff.

Fix by injection rather than environment mutation (mutating `HOME`/`XDG_CONFIG_HOME` in tests is racy under the multithreaded test runner): the loader is split into `load_configs_from(user_config: Option<&Utf8Path>, working_dir)`, and `load_configs` delegates to it with the `dirs` lookup, so public behavior is unchanged. The existing test passes `None`; a new test exercises the user-then-working-dir merge through the injected path, which was previously untestable in isolation.

Verified on a machine with a real user config (which reproduced the failure before the fix): the config tests and the full core suite pass.

---

https://claude.ai/code/session_01EfyrkJkeo3mRngVoppJjUN

### DIFF
--- a/core/src/config/local_fs.rs
+++ b/core/src/config/local_fs.rs
@@ -53,8 +53,7 @@ pub fn load_configs<P: AsRef<Utf8Path>>(working_dir: P) -> Result<Config, Config
     load_configs_from(user_config.as_deref(), working_dir)
 }
 
-/// [`load_configs`] with the user configuration file location injected,
-/// so tests never read the developer's real configuration.
+/// [`load_configs`] with the user configuration file location injected
 pub(crate) fn load_configs_from<P: AsRef<Utf8Path>>(
     user_config: Option<&Utf8Path>,
     working_dir: P,
--- a/core/src/config/local_fs.rs
+++ b/core/src/config/local_fs.rs
@@ -45,14 +45,21 @@ pub fn get_config<P: AsRef<Utf8Path>>(path: P) -> Result<Config, ConfigReadError
 }
 
 pub fn load_configs<P: AsRef<Utf8Path>>(working_dir: P) -> Result<Config, ConfigReadError> {
-    let mut config = dirs::config_dir().map_or_else(
-        || Ok(Config::default()),
-        |mut path| {
-            path.push(CONFIG_DIR);
-            path.push(CONFIG_FILE);
-            get_config(Utf8PathBuf::from_path_buf(path).unwrap())
-        },
-    )?;
+    let user_config = dirs::config_dir().map(|mut path| {
+        path.push(CONFIG_DIR);
+        path.push(CONFIG_FILE);
+        Utf8PathBuf::from_path_buf(path).unwrap()
+    });
+    load_configs_from(user_config.as_deref(), working_dir)
+}
+
+/// [`load_configs`] with the user configuration file location injected,
+/// so tests never read the developer's real configuration.
+pub(crate) fn load_configs_from<P: AsRef<Utf8Path>>(
+    user_config: Option<&Utf8Path>,
+    working_dir: P,
+) -> Result<Config, ConfigReadError> {
+    let mut config = user_config.map_or_else(|| Ok(Config::default()), get_config)?;
     config.merge(get_config(working_dir.as_ref().join(CONFIG_FILE))?);
 
     Ok(config)
--- a/core/src/config/local_fs_tests.rs
+++ b/core/src/config/local_fs_tests.rs
@@ -46,8 +46,7 @@ fn load_configs_merges_user_config_before_working_dir() -> Result<(), Box<dyn Er
         }],
         projects: vec![],
     };
-    wrapfs::File::create(&user_path)?
-        .write_all(toml::to_string_pretty(&user_config)?.as_bytes())?;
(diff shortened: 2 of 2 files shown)


consideRatio:
The prek gate failure was unrelated to this fix: the prek job currently runs the runner image's latest stable clippy (1.97) instead of the pinned 1.96, and its `--fix` hook rewrites four pre-existing needless borrows on any branch cut from main (root cause and durable fix in #454). To keep this PR green standalone, the two one-line-per-file lint commits are included here as well; they are content-identical to the ones on #453, so whichever merges first the other applies as a no-op.

andrius-puksta-sensmetry:
Isn't every integration test that touches indexes or custom sources affected by this? Do they all pass `--no-config`?

consideRatio:
> 🤖 Drafted by Claude Code, reviewed and approved by Erik before posting.

> Isn't every integration test that touches indexes or custom sources affected by this? Do they all pass `--no-config`?

Good instinct to check, but they're already covered — the exposure was narrower than it looks.

The real `~/.config/sysand/sysand.toml` is only reachable through `load_configs` at `sysand/src/lib.rs:177`, which is gated by `--no-config` (it substitutes `Config::default()` and never reads the user config dir). And every CLI integration test spawns the binary through a single helper, `sysand_cmd_in_with` in `sysand/tests/common/mod.rs`, which **unconditionally** injects `--no-config`:

```rust
let args = args
    .into_iter()
    .chain(["--no-config"])
    .chain(cfg.iter().flat_map(|cfg| ["--config-file", cfg]));
```

`cargo_bin!("sysand")` appears only in that helper — no test constructs the command directly (the only other `Command::new` calls spawn `git`). So all the index/source suites (`cli_add_remove`, `cli_source`, `cli_publish`, `cli_lock`, `cli_sync`, `cli_info`, `cli_build`, …) already get `--no-config`, and tests that need config pass it explicitly via `--config-file`. The `core/tests/*` integration tests call the library functions directly and never go through `load_configs` at all.

That's exactly why the unit test was the one exposure: it called the library `load_configs` directly, where there's no `--no-config` escape hatch. `load_configs_from(None, …)` is the library-level equivalent of what the CLI suite already relies on.

One fair follow-up your question surfaces: isolation for the whole CLI suite rests entirely on that single `--no-config` injection, with no `HOME`/`XDG_CONFIG_HOME` sandbox as a backstop — so a future test that bypassed the helper would silently read the real user config. Happy to add temp-dir env sandboxing to the helper as defense-in-depth, but I'd keep it to a separate PR rather than widen this one.

core/src/config/local_fs.rs
```diff
@@ -45,14 +45,21 @@ pub fn get_config<P: AsRef<Utf8Path>>(path: P) -> Result<Config, ConfigReadError
...
+    let user_config = dirs::config_dir().map(|mut path| {
+        path.push(CONFIG_DIR);
+        path.push(CONFIG_FILE);
+        Utf8PathBuf::from_path_buf(path).unwrap()
+    });
+    load_configs_from(user_config.as_deref(), working_dir)
+}
+
+/// [`load_configs`] with the user configuration file location injected,
+/// so tests never read the developer's real configuration.
```
andrius-puksta-sensmetry:
```suggestion
/// [`load_configs`] with the user configuration file location injected
```
Already explained in the test that uses this function.

core/src/config/local_fs_tests.rs
```diff
@@ -25,13 +25,50 @@ fn load_configs() -> Result<(), Box<dyn Error>> {
...
+    let user_path = user_dir.path().join(local_fs::CONFIG_FILE);
+    let user_config = Config {
+        indexes: vec![Index {
+            url: "http://user.example.com".to_string(),
+            ..Default::default()
+        }],
+        projects: vec![],
+    };
+    wrapfs::File::create(&user_path)?
+        .write_all(toml::to_string_pretty(&user_config)?.as_bytes())?;
```
andrius-puksta-sensmetry:
```suggestion
    wrapfs::write(&user_path, toml::to_string(&user_config)?)?;
```

core/src/config/local_fs_tests.rs
```diff
@@ -25,13 +25,50 @@ fn load_configs() -> Result<(), Box<dyn Error>> {
...
+    let working_dir = tempdir()?;
+    let working_config = Config {
+        indexes: vec![Index {
+            url: "http://working.example.com".to_string(),
+            ..Default::default()
+        }],
+        projects: vec![],
+    };
+    wrapfs::File::create(working_dir.path().join(local_fs::CONFIG_FILE))?
+        .write_all(toml::to_string_pretty(&working_config)?.as_bytes())?;
```
andrius-puksta-sensmetry:
```suggestion
    wrapfs::write(working_dir.path().join(local_fs::CONFIG_FILE), toml::to_string(&working_config)?)?;
```

andrius-puksta-sensmetry (approved):

PR #461, opened 2026-07-20
fix(ux): tone down and reduce the amount of warnings about ignored std libs

### LINKED ISSUE #431: Reconsider warnings about std libs
Points to consider/rework about the warnings:
- annoying
- unnecessarily scary (they're informational, so should not be warnings)
- imprecise: in many cases where they're printed they don't actually affect the behaviour/output of the command

Therefore, should consider:
- removing them in favor of command-specific information (e.g. print "standard library usages are ignored" in `sysand info` usages)
- describing in docs why std libs are ignored
- not ignoring std libs for `sysand add - arguably it's better to allow std libs by default

andrius-puksta-sensmetry:
- `add` no longer ignores attempts to add std libs (adding+locking+syncing is now performed like for all other usages); since they will still not be installed during sync, it only informs about that
- `clone` now informs if any (possibly transitive) dependencies will not be installed
- `env install` now has a more specific warning for direct std installation, and warning about std deps becomes a `note`
- `info` now informs about ignored std libs only when they are actually ignored

The only remaining `warnings` about std libs are for `env install`, as that may break other tooling, as explained in the warnings. All other cases are now informational.

Closes #431.

### DIFF
--- a/core/src/commands/add.rs
+++ b/core/src/commands/add.rs
@@ -8,9 +8,8 @@ use crate::{
     },
     project::ProjectMut,
     purl::{PKG_SYSAND_PREFIX, SysandPurlError, parse_sysand_purl},
+    utils::SP,
 };
-
-const SP: char = ' ';
 
 #[derive(Error, Debug)]
 pub enum AddError<ProjectError> {
--- a/core/src/commands/sync.rs
+++ b/core/src/commands/sync.rs
@@ -197,10 +197,8 @@ where
             }
         }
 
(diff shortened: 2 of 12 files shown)


consideRatio:
Re-open to have it trigger on a merge to main as it is now, which includes the prek workflow fix

andrius-puksta-sensmetry:
(will add some tests)

consideRatio (approved):

PR #462, opened 2026-07-20
docs(design): specify v1/whoami in the index API protocol

consideRatio:
Written by erik:

This is a useful endpoint for the client to validate credentials against, where for example using a token for test.sysand.com on sysand.com would otherwise not error on a "sysand auth login" command, because there would be no way to validate it before serious use - and that delay is a bad UX -- better fail early!

Extracted from #453 

Ai text below

## What

Adds a **Token Identity** section to `design/index-api-protocol.md`
specifying the `GET v1/whoami` endpoint:

- bearer authentication
- the `200` body shape: `subject` (type/name semantics per token type)
  and `token` (name, prefix, `expires_at`)
- the unspecified `401` body
- optional rate limiting
- the requirement that the endpoint never consumes single-use credentials

## Why now

The endpoint is already **live on the index server** (sysand.com and
staging), so the protocol spec should be documented in `main` rather than
waiting behind the larger client-side credential-storage work. This PR
extracts just the `index-api-protocol.md` addition from that branch so it
can land on its own.

Spec-only change: no code, no behavior change.

### DIFF
--- a/design/index-api-protocol.md
+++ b/design/index-api-protocol.md
@@ -1,7 +1,8 @@
 # Sysand Index API Protocol
 
 > **Status: partial.** The upload endpoint is described at a high level;
-> the rest of the API surface is not yet specified.
+> the trusted publishing token exchange and `v1/whoami` are specified
+> below; the rest of the API surface is not yet specified.
 
 ## Scope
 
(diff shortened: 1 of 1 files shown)


design/index-api-protocol.md
```diff
@@ -102,3 +103,62 @@ The client discovers `api_root` before attempting trusted publishing.
...
+`subject` identifies the principal the token authenticates as:
+
+- `subject.type` is `user`, `project`, or `oidc`.
+- `subject.name` depends on the token type: for a user token it is the
+  owning user's username; for a project token it is the project id of
+  the scoped project as normalized `<publisher-id>/<name-id>`; for a
+  trusted-publishing (`oidc`) token it is the matched trusted
+  publisher's path (for example the repository full name on GitHub or
+  the project path on GitLab), with a `<provider>:<immutable-id>`
+  fallback when the publisher configuration no longer exists.
```
andrius-puksta-sensmetry, line 145: It would be nice to include the project that the token can publish to.

andrius-puksta-sensmetry, line 145: Are there specific use cases for this (i.e. `subject.name`'s contents)? What to include should be based on that.

I would prefer a more typed approach, but unless the use case requires parsing returned `subject.name`, this is fine.

andrius-puksta-sensmetry: It seems to me like the most useful information for OIDC tokens, as the client has no way to discover this otherwise without actually publishing.

Are there use cases for this? Since trusted publishing tokens are short-lived, I'd say defer decision on the exact format until a use arises.

design/index-api-protocol.md
```diff
@@ -102,3 +103,62 @@ The client discovers `api_root` before attempting trusted publishing.
...
+A missing, malformed, unrecognized, expired, or already-consumed token
+yields 401. The 401 response body is unspecified; clients MUST NOT
+depend on its contents and SHOULD treat any 401 as the credential being
+rejected.
+
+The endpoint is read-only and MUST NOT consume single-use credentials:
+presenting an exchanged trusted-publishing token to `v1/whoami` leaves
+it usable for the subsequent upload. Servers MAY rate limit the
+endpoint (429), since it can be probed without a valid credential and
+acts as a token-validity oracle.
```
andrius-puksta-sensmetry:
```suggestion
it usable for the subsequent upload. Servers SHOULD rate limit the
endpoint (429), since it can be probed without a valid credential and
acts as a token-validity oracle.
```
I'd say the rate limit should be set very low, like 1 in 30 seconds, 10 an hour, 15 a day.

design/index-api-protocol.md
consideRatio: I deliberated back and forth about this, ended up not going for it to avoid complexity for something non-essential.

For users it would be a different sql query than for project tokens and oidc etc.
Currently we just have publish or not permissions, but you can imagine different permissions in the future, making this endpoint then in need of updates and potentially with breaking change communication if we would ever depend on its structure in the client etc.

I lean towards waiting to add it until we find it relevant enough to re-evaliate after using it a while.

consideRatio: The absolute essential use case is valid / not-valid at all
The bonus aspect is getting some insight about the configured token, to refresh memory about what was once configured (now that you persist creds over time).

Currently, the clients sysand auth whoami command in #453 is just presenting string values under these keys.

andrius-puksta-sensmetry (approved):
Good enough for now.

> using a token for test.sysand.com on sysand.com would otherwise not error on a "sysand auth login" command, because there would be no way to validate it before serious use

~~Token can e.g. be checked with e.g. `HEAD` request for `index.json`.~~
Edit: nevermind, public indexes don't require a token for reading...

Issue #437, opened 2026-07-06
Explore options to managing credentials to indexes


consideRatio:
The experience on configuring auth for a private index isn't great, and I have no fix proposed. Instead this is an issue about exploring options to potentially develop an improvement.

The background is that when you have a token to access an index, its really not clear how we propose users manage the token to let sysand get access in a user friendly way over time. They could configure the env for a specific session, but tokens they get are typically presented only once --- so how should users configure it? If they configure it once in a session, a hacker may even extract it from their computer because of shell history mistakes, while the user may fail to re-acquire it in another terminal session.

Its a problematic UX we currently have around this, and its also problematic to have a bad UX as the index tokens could be for example "read all the users repos" on gitlab currently, which is very sensitive.

andrius-puksta-sensmetry:
One option is to copy [`cargo login` behavior](https://doc.rust-lang.org/cargo/commands/cargo-login.html), which stores credentials in either `~/.cargo/credentials.toml` or using other credential providers (e.g. OS keyring).

andrius-puksta-sensmetry:
It is a common practice (not a good one, but no good universal alternative exists) to store credentials in plain text by default. This is done by e.g.:
- web browsers (AFAIK Firefox does encrypt them, but with a static key that can be extracted from the firefox binary)
- ssh
- cargo

Password-based encryption can also be used as e.g. is recommended for SSH keys. We could use that, but without integration with a credential helper, the password would need to be typed every time sysand reads that credential from disk (which it would for most commands).

consideRatio:
## Keyring compatibility across the three OSes

The Rust `keyring` crate (v3) is a thin wrapper over each OS's native secret store. The three platforms are not equally reliable, and Linux is the whole reason a fallback exists.

| Platform | Backend | Daemon / session needed? | Headless / CI / container | Reliability estimate |
|---|---|---|---|---|
| macOS | Keychain Services (Security.framework) | No daemon; uses the login keychain | Works over local session; can fail on pure SSH with no GUI to show the unlock/allow prompt | ~95%+. Effectively always works for a logged-in user |
| Windows | Credential Manager (Win32 `CredRead`/`CredWrite`) | No daemon; per-user vault, present on desktop and Server | Works headless too (per-user vault exists without a desktop) | ~99%. The most trouble-free of the three. One caveat: blob limit ~2560 bytes, irrelevant for tokens |
[...]

PR #478, opened 2026-07-24
feat(discovery)!: require explicit api_root for an index API

consideRatio:
## Summary

Stop defaulting `api_root` to the discovery root. An index now exposes an API **only** when its discovery document (`sysand-index-config.json`) advertises `api_root` explicitly. Absent that, the index is read-only, regardless of whether its discovery root is a plain URL or a URL template.

This makes the invariant clean and unambiguous:

> "the index has an API" == "the discovery document advertises `api_root`"

Previously a plain-URL index implicitly gained an `api_root` equal to its discovery root, so it *looked* like it had an API surface even when it was a static-file mirror. That ambiguity is exactly what makes it hard, elsewhere in the client, to tell "I failed to reach the API" apart from "there is no API". Removing the implicit default resolves it at the source.

## Why now

This is the decoupled prerequisite for the credential-storage work (#453 / #437): login validation only wants to probe an API surface when one genuinely exists. With this landed, "advertises `api_root`" becomes the single reliable signal for "has an API", and the credential-storage branch can rely on it.

## Behavior changes

- **Reads**: unaffected. Index reads use `index_root`, which still defaults to the discovery root.
- **Publish**: publishing to an index that advertises no `api_root` now fails early with `does not advertise a publish endpoint` instead of composing an upload request against the discovery root. (This message already existed for templated indexes; it is now the general case.)
- **Pasting the upload URL as `--index`**: now yields the same "no publish endpoint" refusal (404 discovery → no `api_root`) rather than reaching `api_root` shape validation. Shape validation still guards an *advertised* `api_root` that points at the upload endpoint, and keeps its unit coverage.

**BREAKING CHANGE**: a plain-URL index that previously accepted uploads at the discovery root must now advertise `api_root` in `sysand-index-config.json`.

## Spec

`design/index-protocol.md` and `design/index-api-protocol.md` updated: `api_root` has no default; absence means no API / read-only.

## Tests

- New: `discovery_absent_config_yields_no_api_root`, `discovery_config_without_api_root_yields_no_api_root` lock the new contract (404 config and `{}` config both leave `api_root` unset).
- Repurposed `publish_rejects_upload_endpoint_index_url` to assert the new bail path.
- Full `sysand-core --all-features` lib suite (352) and the entire `sysand` CLI integration suite pass; clippy clean across core + cli all-features.

### DIFF
--- a/core/src/commands/publish.rs
+++ b/core/src/commands/publish.rs
@@ -295,16 +295,6 @@ pub fn validate_api_root_url_shape(url: &Url) -> Result<(), PublishError> {
     if url.fragment().is_some() {
         return Err(err("URL must not include a fragment component".to_string()));
     }
-    // Reject a URL that already names the upload endpoint. Catches the
-    // common mistake of pasting the full upload URL into `--index`,
-    // which would otherwise either compose to `v1/upload/v1/upload`
-    // (after discovery defaulted `api_root` to the discovery root) or
-    // send a discovery request to a path that can never serve one.
-    if url.path().trim_end_matches('/').ends_with("v1/upload") {
-        return Err(err(
-            "URL must be a discovery root or `api_root`, not the `v1/upload` endpoint".to_string(),
-        ));
-    }
     Ok(())
 }
 
--- a/sysand/src/cli.rs
+++ b/sysand/src/cli.rs
@@ -210,11 +210,8 @@ pub enum Command {
         path: Option<Utf8PathBuf>,
 
         /// Configured index URL to publish to (e.g. https://sysand.com)
-        /// May point to a path containing sysand-index-config.json, or directly
-        /// to the API root (e.g. https://sysand.com/api)
-        /// URL templates (see --index under resolution options) are accepted,
-        /// but publishing then requires the index's sysand-index-config.json
-        /// to set `api_root`
+        /// The index must advertise a publish endpoint: its
+        /// sysand-index-config.json must set `api_root`
         #[arg(long, value_name = "URL", verbatim_doc_comment)]
         index: IndexLocation,
 
--- a/core/src/commands/publish_tests.rs
+++ b/core/src/commands/publish_tests.rs
@@ -563,22 +563,6 @@ fn validate_api_root_rejects_non_http_scheme() {
 }
 
 #[test]
-fn validate_api_root_rejects_upload_endpoint_path() {
-    // An `api_root` that already ends in the upload path means the full
-    // upload URL was pasted in the wrong place; catch it before we would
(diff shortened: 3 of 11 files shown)


core/src/commands/publish.rs
```diff
@@ -295,11 +295,10 @@ pub fn validate_api_root_url_shape(url: &Url) -> Result<(), PublishError> {
...
-    // which would otherwise either compose to `v1/upload/v1/upload`
-    // (after discovery defaulted `api_root` to the discovery root) or
-    // send a discovery request to a path that can never serve one.
+    // Reject a URL that already names the upload endpoint. Catches an
+    // `api_root` advertised by the discovery document that points at the
+    // upload endpoint itself, which would otherwise compose to
+    // `v1/upload/v1/upload`.
     if url.path().trim_end_matches('/').ends_with("v1/upload") {
         return Err(err(
             "URL must be a discovery root or `api_root`, not the `v1/upload` endpoint".to_string(),
```
andrius-puksta-sensmetry: Remove this check.

#453 adds `api/v1/whoami`, so the API no longer serves a single endpoint, making this error highly unlikely. Since this function now checks exclusively the server-advertised `api_root` (i.e. CLI-supplied values can no longer reach this check), ensuring the correct configuration becomes the responsibility of the server operator, and IMO not worth checking by the client.

sysand/src/cli.rs
```diff
@@ -210,11 +210,10 @@ pub enum Command {
...
         /// Configured index URL to publish to (e.g. https://sysand.com)
-        /// May point to a path containing sysand-index-config.json, or directly
-        /// to the API root (e.g. https://sysand.com/api)
-        /// URL templates (see --index under resolution options) are accepted,
-        /// but publishing then requires the index's sysand-index-config.json
-        /// to set `api_root`
+        /// The index must advertise a publish endpoint: its
+        /// sysand-index-config.json must set `api_root`. This is required
+        /// for any URL shape, including URL templates (see --index under
+        /// resolution options)
```
```suggestion
        /// The index must advertise a publish endpoint: its
        /// sysand-index-config.json must set `api_root`
```
`sysand publish --help` does not have `Resolution options` section, so let's not confuse users. Also there is no reason for users to believe that templated URLs are treated differently here.

andrius-puksta-sensmetry (changes requested):
Good, it was getting rather confusing with the templates and the upcoming `/whoami` special treatment.

Just a couple nits.

consideRatio:
Asked AI to address review and rebase etc, havnt inspected this myself

andrius-puksta-sensmetry (approved):

PR #488, opened 2026-08-04
fix: set zip date to 1980-01-02, fix CI and test feature gating

andrius-puksta-sensmetry:
`cargo test -p sysand-core` did not work due to references to `keyring` crate. CI was not actually `cargo check`ing all feature combinations, since due to feature unification (with CLI crate) most features of core were always enabled.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -76,20 +76,30 @@ jobs:
 
       # cargo clippy is run with variants of features enabled, where
       # --all-features is done by prek so we don't do it here
+      # To avoid feature unification (which enables the superset of features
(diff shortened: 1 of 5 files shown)


consideRatio:
Hmmm, 2026-01-01 could be indicative of a real date - while 1980 couldn't. What do you think about 1981 or 1980-01-02 instead?

consideRatio (approved):
We can proceed, but I think it is a UX risk to declare 2026 01 01 as it can be confused with an actual date something was set, and for example make people think -- oh this is not updated so recently, instead of thinking "oh this is a dummy date"

PR #491, opened 2026-08-04
feat: preserve datetime of building the KPAR in the zip metadata

andrius-puksta-sensmetry:
We initialy wanted to make kpars reproducible, so set datetime to earliest possible (#372). To prevent confusion of users who download and unzip a kpar manually (and then see that the files date back to 1980), set the datetime to the build datetime (same for all files). If we find a good reason to have reproducible kpars, this can be changed in the future.

What do you think @consideRatio? We don't currently have good reasons to do either way, but having the correct time is less confusing to users.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5529,6 +5529,7 @@ checksum = "2d04a6b5381502aa6087c94c669499eb1602eb9c5e8198e534de571f7154809b"
 dependencies = [
  "aes 0.9.1",
  "bzip2",
+ "chrono",
  "constant_time_eq",
  "crc32fast",
  "deflate64",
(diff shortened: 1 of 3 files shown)


consideRatio (approved):
I think this is OK to do, not formed a clear opinion yet.

It seems Python wheels built with `wheel` is letting source file's modified time be respected in the .whl files (zip), but `pip` when installing a wheel, will install content so its declared modified as "now".

So we have the following considerations:
1. What .kpar / zip do we write -- perhaps 1980 / now / last-modified
2. What .kpar / zip files do we unpack when using sysand -- perhaps 1980 / now / archive's "modified" info

andrius-puksta-sensmetry:
Regarding 2., I'd say it's fine to set it to now (we currently do this, although not explicitly - we just create a file and write to it, so its last modified time is set by the OS).

Regarding 1., file's last modified time could be used, but it would involve an additional syscall for every file, and IMO would not provide any benefit.

PR #495, opened 2026-08-10
chore(bindings): update `jni` to 0.22, explicitly register Java methods

andrius-puksta-sensmetry:
Also initialize the logger for Java bindings, previously it was not initialized and any log calls in core were no-ops when called from Java bindings.

### DIFF
--- a/macros/Cargo.toml
+++ b/macros/Cargo.toml
@@ -13,7 +13,7 @@ repository.workspace = true
 proc-macro = true
 
 [dependencies]
-itertools = { version = "0.14.0", default-features = false  }
+itertools = { version = "0.15.0", default-features = false  }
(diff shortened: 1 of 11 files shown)


macros/Cargo.toml
```diff
@@ -16,4 +16,4 @@ proc-macro = true
 itertools = { version = "0.14.0", default-features = false  }
 proc-macro2 = "1.0.106"
 quote = "1.0.45"
-syn = "2.0.117"
+syn = "3.0"
```
andrius-puksta-sensmetry, line 19: This is unrelated, but did not require any changes.

vytautas-astrauskas-sensmetry (approved):

PR #526, opened 2026-09-07
test(py): add `mock_index` fixture and a red end-to-end scenario suite

vytautas-astrauskas-sensmetry:
This pull request adds test infrastructure for the Python bindings: an
in-process sysand index served by `pytest-httpserver` and a factory that builds
a real project against it.

The only change to shipped code is that `_run_cli` now releases the GIL
while the CLI runs.

## Why

We need to extend the Python bindings to cover full Sysand functionality, which
requires us to be able to test it against a live index from Python. Until now
the only index-facing Python test hand-wrote `httpserver` expectations for a
single lookup. Testing the new functions that way would mean duplicating the
index layout in every test and would leave the interesting cases (a version
published mid-test, a 401, a 500 on one archive) untested.

As a motivating example for the testing infrastructure and new Python API is
using Python API to update a Sysand package.

### `tests/mockindex.py` — `MockIndex`

Serves the real index protocol layout over HTTP, mirroring
`core/src/env/discovery.rs`:

- `pkg:sysand/<publisher>/<name>` at `/<publisher>/<name>/`, any other IRI
  at `/_iri/<sha256hex>/`;
- per project `versions.json` (kept in strictly descending semver order, as
  ingest validation requires) and per version `.project.json`, `.meta.json`
  and `project.kpar` with a reproducible archive digest;
- `/sysand-index-config.json` answers 404, so the index root defaults to the
  discovery root.

Nothing on the sysand side is stubbed: the CLI and the bindings talk to the
server exactly as to a live index.

Test controls: `publish()` may be called mid-test; `require_bearer(token)`
turns every unauthenticated request into a 401; `fail_next(path, status)`
injects one failure; `requests(glob)` returns the request log.
`run_cli_in(root, *args)` runs the in-process CLI with `root` as the
working directory and restores the previous one afterwards.

### `tests/conftest.py` — fixtures

- `mock_index`: wraps `MockIndex` and, on teardown, fails the test if a
  handler raised or if `/index.json` was fetched. Lookups must go through
  the per-project `versions.json`, never enumerate the index; this is the
  same invariant `sysand/tests/cli_lock.rs` checks.
- `make_baseline` / `baseline`: publishes `pkg:sysand/mock/library` 0.10.3, runs
  `sysand.init`, writes a manifest by hand and then runs `lock` and `sync`
  through the shipped CLI.
- `_sysand_isolation` (autouse): points the credential store at the
  debug-only `:absent:` seam and clears every `SYSAND_*` variable,
  so tests never touch the developer's keyring or configuration. This
  mirrors `sysand/tests/common/mod.rs`, which is also updated to clear all
  `SYSAND_` env vars.
- `_claim_logger` (session, autouse): makes one binding call before any
  test so `pyo3_log` owns the global logger. Otherwise the first `_run_cli`
  call could install `env_logger` instead and the tests that assert on
  `caplog` records would see nothing.

### `tests/test_mock_index.py`

The fixture's own tests: layout paths for PURL and non-PURL IRIs, duplicate
and non-semver publish rejection, bearer enforcement, fault injection, cwd
restoration, and `lock` + `sync` through the CLI against the mock index.
The last one is the Python twin of
`sysand/tests/cli_lock.rs::lock_and_sync_against_mock_index`.

### `test_basic.py::test_index_info`

Rewritten on top of `mock_index` instead of hand-written `httpserver`
expectations. The assertions are unchanged; it additionally checks that
only the project's `versions.json` was requested.

### `src/lib.rs` — `_run_cli` releases the GIL

`_run_cli` now runs the CLI inside `py.detach(...)`. The CLI can run for
seconds and talk to the network. Holding the GIL for that long blocked
every other Python thread, including `pytest-httpserver`'s thread, so a
CLI call against the in-process index deadlocked. Independently of tests,
a long-running CLI call should not freeze the interpreter.

### DIFF
--- a/bindings/py/tests/conftest.py
+++ b/bindings/py/tests/conftest.py
@@ -1,0 +1,162 @@
+# SPDX-License-Identifier: MIT OR Apache-2.0
+# SPDX-FileCopyrightText: © 2026 Sysand contributors <opensource@sensmetry.com>
+
+from __future__ import annotations
+
+import dataclasses
+import json
+import os
+import typing
+from pathlib import Path
+
+import pytest
+from pytest_httpserver import HTTPServer
+
+import sysand
+from mockindex import (
+    LEGACY_CONSTRAINT,
+    LIBRARY,
+    MockIndex,
+    run_cli_in,
+    usage,
+)
+
+
+@pytest.fixture(scope="session", autouse=True)
+def _claim_logger(tmp_path_factory: pytest.TempPathFactory) -> None:
+    # Whoever installs the global `log` logger first owns it for the process:
+    # binding functions install `pyo3_log` (records reach `caplog`), while
+    # `_run_cli` tries `env_logger`. Make one cheap binding call before any
+    # test so `_run_cli` can never steal the logger from the tests that
+    # assert on log records; `run_cli` then only prints a "failed to set up
+    # logger" warning and continues.
+    sysand.env.env(tmp_path_factory.mktemp("logger") / sysand.env.DEFAULT_ENV_NAME)
+
+
+def isolate_sysand_env(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Mirrors `sysand/tests/common/mod.rs`.
+
+    Drop every ambient `SYSAND_*` override (index, config, credentials, and
+    whatever is added later) so tests see only what they set themselves, then
+    point the credential store at the debug-only seam in
+    `sysand/src/credential_store.rs` so the developer's real OS keyring is
+    never touched.
+    """
+    for var in [v for v in os.environ if v.startswith("SYSAND_")]:
+        monkeypatch.delenv(var)
+    monkeypatch.setenv("SYSAND_TEST_CREDENTIAL_STORE", ":absent:")
+    monkeypatch.setenv("NO_COLOR", "1")
+
+
+@pytest.fixture(autouse=True)
+def _sysand_isolation(monkeypatch: pytest.MonkeyPatch) -> None:
+    isolate_sysand_env(monkeypatch)
+
+
+@pytest.fixture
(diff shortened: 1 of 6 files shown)


bindings/py/tests/conftest.py
```diff
@@ -0,0 +1,160 @@
...
+    monkeypatch.setenv("NO_COLOR", "1")
+    for var in (
+        "SYSAND_DEFAULT_INDEX",
+        "SYSAND_INDEX",
+        "SYSAND_CONFIG_FILE",
+        "SYSAND_NO_CONFIG",
+    ):
+        monkeypatch.delenv(var, raising=False)
+    for var in [v for v in os.environ if v.startswith("SYSAND_CRED_")]:
+        monkeypatch.delenv(var)
```
andrius-puksta-sensmetry:
```suggestion
    # Mirrors `sysand/tests/common/mod.rs`: never touch the developer's real
    # OS keyring (debug-only seam in `sysand/src/credential_store.rs`), and
    # keep any other sysand env vars out
    for var in [v for v in os.environ if v.startswith("SYSAND_")]:
        monkeypatch.delenv(var)
    monkeypatch.setenv("SYSAND_TEST_CREDENTIAL_STORE", ":absent:")
    monkeypatch.setenv("NO_COLOR", "1")
```
Or something to this effect; we may add more env vars later, and they will (almost certainly) use `SYSAND_` prefix, so delete them all to avoid any issues.

andrius-puksta-sensmetry: `sysand/tests/common/mod.rs` should also be changed accordingly.

vytautas-astrauskas-sensmetry: Fixed in f76c9a09502417837620416d08b1e2580e39e4fb

andrius-puksta-sensmetry (approved):
LGTM

PR #527, opened 2026-09-08
feat(py): lossless `set_usage_constraint`, `add` returns its bool

vytautas-astrauskas-sensmetry:
This MR adds `sysand.set_usage_constraint(path, resource, constraint, *, must_exist=True)`, which replaces (or clears) the version constraint of one usage in `.project.json` without disturbing anything else in the file, and makes `sysand.add` return whether it actually added a usage. It also introduces `sysand.SysandError` and `sysand.ProjectError`, the first two classes of a typed exception hierarchy for the bindings.

## Why

A tool that upgrades a dependency on the user's behalf needs to move one usage from, say, `>=0.10.0, <0.11.0` to `>=0.11.0, <0.12.0` and show the user what it did. The bindings offered no way to do that well:

- `add` on an already declared usage does not replace the constraint. It merges the new requirement into the existing one, so the two ranges above become an unsatisfiable conjunction.
- Every typed read-modify-write of `.project.json` (`add`, `remove`) goes through the Rust model, which drops keys it does not know and rewrites the remaining keys in its own order. A tool that edits a manifest owned by a user, possibly carrying extension keys, cannot afford that.
- `add` already computed whether a usage was added, but both the Rust binding and the Python wrapper discarded the value, so a caller could not tell "declared it" from "merged into an existing declaration".
- Failures surfaced as a bare `RuntimeError`, giving a caller no way to distinguish bad input from a broken project, or to know whether anything had been written before the failure.

### `core/src/commands/usage.rs` — `do_set_usage_constraint`

Edits the parsed JSON document rather than the typed model. `serde_json` is already built with `preserve_order`, so every other key, unknown keys included, and the key order survive verbatim. Only the matched usage's `versionConstraint` value is touched: it is replaced in place, appended when the usage had none, or removed (order-preserving) when clearing.

Validation happens before any modification: the constraint must parse as a semver requirement, and a resource declared more than once is refused as `Ambiguous` instead of guessing. The resource is matched the way `add` matches it: the `publisher/name` shorthand is expanded to its `pkg:sysand/` PURL, then compared as a plain string. Directory and kpar-path usages carry no `resource` key and never match.

The result is `(matched_resource, ConstraintChange)`, where the change is `Replaced { old, new }`, `Unchanged { constraint }` or `NotFound`, so callers can report precisely what happened. `do_set_usage_constraint_local` applies it to a `LocalSrcProject`.

### `core/src/project/local_src.rs` — `edit_info_document`

A new method on `LocalSrcProject` that reads `.project.json` as a document, applies a caller-supplied edit, and writes the result back in the same byte shape `put_info` produces (pretty-printed, trailing newline), but only if the document changed. The whole file is serialized to memory first, so a failing edit or serialization never truncates the manifest. Errors are `EditInfoError<E>`: either the project could not be read or written, or the edit itself failed with `E`.

### Python: `set_usage_constraint`, `UsageConstraintChange`

`sysand.set_usage_constraint` returns a `UsageConstraintChange` TypedDict with `resource` (the expanded IRI actually matched), `found`, `changed`, `old_constraint` and `new_constraint`. With `must_exist=True` (the default) a missing usage raises `ProjectError`; with `False` it is reported as `found=False` and nothing is written. Bad input (an invalid requirement or a malformed shorthand) raises `ValueError`; a missing, malformed or ambiguous manifest raises `ProjectError`.

Because whitespace is normalised to sysand's own format, a manifest last written by sysand changes in exactly one line. A hand-formatted manifest gets sysand's whitespace on first edit but keeps all of its keys and their order.

### Python: `SysandError`, `ProjectError`

Defined in `sysand/_errors.py` and raised from Rust via `pyo3::import_exception!`, so `mypy` sees their attributes. `SysandError` subclasses `RuntimeError`, so existing `except RuntimeError` handlers keep working, and carries `wrote: bool`, which tells a caller whether the failed call modified anything on disk. Every error raised in this MR has `wrote=False`: nothing is written unless the edit succeeds.

### Python: `add` returns `bool`

`do_add_py` now passes through the value `do_add_guess` already produced, and `sysand.add` returns it: `True` when a new usage was added, `False` when the call merged into an existing declaration.

### Tests

- `core/src/commands/usage_tests.rs`: document-level tests, including a one-line-diff check on a manifest with unknown keys and non-canonical order, shorthand matching, clear keeps order, append when absent, `null` counts as absent, unchanged leaves the document alone, not-found cases, directory usages never match, and every refusal (ambiguous, invalid constraint, malformed shorthand, malformed document) happens before any edit.
- `bindings/py/tests/test_basic.py`: pytest equivalents on real files, plus "unchanged does not touch the file" (mtime and bytes), hand-formatted whitespace normalisation, and the `add` return value.
- `bindings/py/tests/test_migration_scenarios.py`: two end-to-end tests against the `baseline` fixture from the mock index. One moves the library's constraint and asserts the manifest differs in exactly one line with extension keys and order intact; the other covers a project that does not declare the library at all (`must_exist=False`, the refusal with `wrote=False`, then `add` declaring it and a second `add` merging).
- `bindings/py/tests/conftest.py`: the baseline manifest is now written in sysand's own two-space format, which is the documented scope of the one-line guarantee.

### DIFF
--- a/bindings/py/tests/test_basic.py
+++ b/bindings/py/tests/test_basic.py
@@ -79,6 +79,186 @@ def test_remove_accepts_sysand_shorthand() -> None:
         )
 
 
+SYSAND_FORMATTED_MANIFEST = """{
+  "version": "1.2.3",
+  "name": "fidelity",
+  "x-unknown-top-level": {
+    "kept": true
+  },
+  "publisher": "acme",
+  "usage": [
+    {
+      "dir": "../local-lib",
+      "publisher": "local-pub",
+      "name": "local-lib"
+    },
+    {
+      "x-unknown-in-usage": null,
+      "resource": "pkg:sysand/mock/library",
+      "versionConstraint": ">=0.10.0, <0.11.0"
+    }
+  ]
+}
+"""
+
+
+def _one_line_diff(before: str, after: str) -> tuple[str, str]:
+    old_lines = before.splitlines(keepends=True)
+    new_lines = after.splitlines(keepends=True)
+    assert len(old_lines) == len(new_lines)
+    differing = [(o, n) for o, n in zip(old_lines, new_lines) if o != n]
+    assert len(differing) == 1, differing
+    return differing[0]
+
+
+def test_set_usage_constraint_preserves_document(tmp_path: Path) -> None:
+    manifest = tmp_path / ".project.json"
+    manifest.write_text(SYSAND_FORMATTED_MANIFEST)
+
+    change = sysand.set_usage_constraint(tmp_path, "mock/library", ">=0.11.0, <0.12.0")
+
+    assert change == {
+        "resource": "pkg:sysand/mock/library",
+        "found": True,
+        "changed": True,
+        "old_constraint": ">=0.10.0, <0.11.0",
+        "new_constraint": ">=0.11.0, <0.12.0",
+    }
+    old_line, new_line = _one_line_diff(SYSAND_FORMATTED_MANIFEST, manifest.read_text())
+    assert old_line == '      "versionConstraint": ">=0.10.0, <0.11.0"\n'
+    assert new_line == '      "versionConstraint": ">=0.11.0, <0.12.0"\n'
+
+
+def test_set_usage_constraint_purl_matches_shorthand_declaration(
+    tmp_path: Path,
+) -> None:
+    (tmp_path / ".project.json").write_text(SYSAND_FORMATTED_MANIFEST)
+    change = sysand.set_usage_constraint(tmp_path, "pkg:sysand/mock/library", "0.11.0")
+    assert change["found"] and change["changed"]
+    assert change["new_constraint"] == "0.11.0"
+
+
+def test_set_usage_constraint_clear(tmp_path: Path) -> None:
+    manifest = tmp_path / ".project.json"
+    manifest.write_text(SYSAND_FORMATTED_MANIFEST)
+
+    change = sysand.set_usage_constraint(tmp_path, "mock/library", None)
+
+    assert change["changed"] is True
+    assert change["old_constraint"] == ">=0.10.0, <0.11.0"
+    assert change["new_constraint"] is None
+    expected = SYSAND_FORMATTED_MANIFEST.replace(
+        ',\n      "versionConstraint": ">=0.10.0, <0.11.0"', ""
+    )
+    assert manifest.read_text() == expected
+
+    # Setting a constraint on a usage without one appends it as the last key.
+    sysand.set_usage_constraint(tmp_path, "mock/library", "^0.11")
+    assert manifest.read_text() == SYSAND_FORMATTED_MANIFEST.replace(
+        ">=0.10.0, <0.11.0", "^0.11"
+    )
+
+
+def test_set_usage_constraint_unchanged_does_not_touch_file(tmp_path: Path) -> None:
+    manifest = tmp_path / ".project.json"
+    manifest.write_text(SYSAND_FORMATTED_MANIFEST)
+    os.utime(manifest, ns=(1_000_000_000_000_000_000, 1_000_000_000_000_000_000))
+    before = manifest.stat()
+
+    change = sysand.set_usage_constraint(tmp_path, "mock/library", ">=0.10.0, <0.11.0")
+
+    assert change["found"] is True
+    assert change["changed"] is False
+    assert change["old_constraint"] == change["new_constraint"] == ">=0.10.0, <0.11.0"
+    after = manifest.stat()
+    assert after.st_mtime_ns == before.st_mtime_ns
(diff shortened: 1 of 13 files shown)


core/src/commands/usage.rs
```diff
@@ -0,0 +1,164 @@
...
+    /// The document was not modified.
+    Unchanged { constraint: Option<String> },
+    /// No usage matched the resource. The document was not modified.
+    NotFound,
+}
+
+#[derive(Debug, Error)]
+pub enum SetConstraintError {
+    #[error("`{0}` is not a valid version requirement: {1}")]
+    InvalidConstraint(String, #[source] semver::Error),
```
andrius-puksta-sensmetry:
```suggestion
    #[error("`{0}` is not a valid version requirement")]
    InvalidConstraint(String, #[source] semver::Error),
```
Sources chain is formatted separately.

core/src/commands/usage_tests.rs
```diff
@@ -0,0 +1,339 @@
...
+    let (_, change) = do_set_usage_constraint(&mut d, "local-pub/local-lib", Some("1")).unwrap();
+    assert_eq!(change, ConstraintChange::NotFound);
+}
+
+#[test]
+fn ambiguous_is_refused_before_editing() {
+    let mut d = doc(&MANIFEST.replace(
+        r#""dir": "../local-lib""#,
+        r#""resource": "pkg:sysand/mock/library""#,
+    ));
```
andrius-puksta-sensmetry: Tangent: somewhat surprisingly, replacing `dir: ...` with `resource: ...` does cause the usage to parse as `Resource`, even though it still contains publisher and name. Part of a broader issue of ignoring unknown fields, which was worked around in the previous PR, but should be properly fixed. Opened #530 for that.

core/src/commands/usage_tests.rs
```diff
@@ -0,0 +1,339 @@
...
+
+        do_set_usage_constraint_local(&mut project, "a:b", Some("2.0.0")).unwrap();
+
+        assert_eq!(
+            std::fs::read_to_string(&path).unwrap(),
+            concat!(
+                "{\n  \"name\": \"hand\",\n  \"version\": \"1.0.0\",\n  \"usage\": [\n    {\n",
+                "      \"resource\": \"a:b\",\n      \"versionConstraint\": \"2.0.0\"\n    }\n",
+                "  ],\n  \"z\": 1\n}\n"
+            )
```
andrius-puksta-sensmetry: Unreadable. Use a raw string for this.

bindings/py/tests/test_basic.py
```diff
@@ -79,6 +79,176 @@ def test_remove_accepts_sysand_shorthand() -> None:
...
+    manifest = tmp_path / ".project.json"
+    manifest.write_text(
+        '{"name":"hand","version":"1.0.0","usage":[{"resource":"a:b","versionConstraint":"1.0.0"}],"z":1}'
+    )
+    sysand.set_usage_constraint(tmp_path, "a:b", "2.0.0")
+    # Same keys, same order, sysand's whitespace.
+    assert manifest.read_text() == (
+        '{\n  "name": "hand",\n  "version": "1.0.0",\n  "usage": [\n    {\n'
+        '      "resource": "a:b",\n      "versionConstraint": "2.0.0"\n    }\n'
+        '  ],\n  "z": 1\n}\n'
```
andrius-puksta-sensmetry: Use `"""` string for readability.

andrius-puksta-sensmetry (approved):
Good to go when the nits above are fixed.

core/src/commands/usage.rs
vytautas-astrauskas-sensmetry: Fixed.

core/src/commands/usage_tests.rs
vytautas-astrauskas-sensmetry: Replaced with:

```rust
    let mut d = doc(MANIFEST);
    d["usage"][0] = serde_json::json!({ "resource": LIBRARY });
```

core/src/commands/usage_tests.rs
vytautas-astrauskas-sensmetry: Replaced with:

```rust
        assert_eq!(
            std::fs::read_to_string(&path).unwrap(),
            r#"{
  "name": "hand",
  "version": "1.0.0",
  "usage": [
    {
      "resource": "a:b",
      "versionConstraint": "2.0.0"
    }
  ],
  "z": 1
}
"#
        );
    }
```

bindings/py/tests/test_basic.py
vytautas-astrauskas-sensmetry: Replaced with:

```python
    assert (
        manifest.read_text()
        == """{
  "name": "hand",
  "version": "1.0.0",
  "usage": [
    {
      "resource": "a:b",
      "versionConstraint": "2.0.0"
    }
  ],
  "z": 1
}
"""
    )
```

PR #528, opened 2026-09-08
feat(py): `env.projects`, `discover`, export `env.sources`

vytautas-astrauskas-sensmetry:
# feat(py): `env.projects`, `discover`, export `env.sources`

This MR lets a Python caller inspect the state the CLI otherwise keeps to itself: what is installed in a `.sysand` environment, and which project and workspace enclose a given path.

- `sysand.env.projects(env_path)` returns the entries of the environment's `env.toml` as typed dicts.
- `sysand.discover(path)` returns the enclosing project root and workspace root, or `None` for each, using the same lookup every CLI command starts with.
- `sysand.env.sources` is exported. The module existed but was never added to `sysand.env.__all__`, so the environment-side listing of a project's sources was unreachable through the public API.
- A new `EnvError` (a `SysandError`, hence a `RuntimeError`) replaces the bare `IOError` that was raised when an environment could not be read.

## Why

A tool that drives sysand from Python needs to answer two questions after `sync` that the bindings could not answer:

- Which projects are installed now, where, and under which IRIs? `env.toml` is an internal format. Parsing it from Python would couple the caller to its layout and to the path conventions of editable versus copied entries, which are exactly what `LocalDirectoryEnvironment` already encapsulates.
- Am I inside a project, and is that project part of a workspace? `sysand.root` finds the project root but says nothing about workspaces, and the CLI's workspace detection lived only behind the CLI. A tool that must refuse to operate inside a workspace had no way to check.

Both are read-only lookups over code the CLI already runs, so exposing them is a matter of wiring rather than new behaviour.

### `sysand.env.projects` (`env/_projects.py`, `lib.rs::do_env_projects_py`)

Reads the environment with `LocalDirectoryEnvironment::read` and maps each entry of `env.toml` to an `EnvProject` typed dict:

| field | meaning |
|---|---|
| `publisher`, `name`, `version` | as recorded in `env.toml` |
| `path` | verbatim from `env.toml`: relative to the environment directory, or to the workspace/project root for `editable` entries. Not joined. |
| `identifiers` | IRIs of the project, the first canonical. Empty only for `editable` entries. |
| `usages` | IRIs of the project's usages |
| `editable`, `workspace` | the two flags from `env.toml` |

`path` is deliberately left relative, because the base differs by entry kind. The docstring on `EnvProject` says which base to join it onto, and the test covers both cases (`os.path.samefile(root / own["path"], root)` for the editable entry, `env_dir / dep["path"]` for copied ones).

The Rust side returns plain tuples and the Python side builds the dicts, the same split the other bindings use.

### `sysand.discover` (`_discover.py`, `lib.rs::do_discover_py`)

Calls `discover_project` and `discover_workspace` from `sysand_core::discover`, so the result is the same the CLI computes before running a command: walk up from `path` to the nearest directory holding `.project.json` or `.meta.json`, and separately to the nearest one holding `.workspace.json`. Both paths are canonicalized as `sysand.root` does, and `Discovery` is a typed dict with `project_root` and `workspace_root`, each `Optional[str]`. A directory that cannot be read or a malformed `.workspace.json` raises `ProjectError` with `wrote == False`.

`sysand.root` is unchanged; `discover(...)["project_root"]` equals `str(root(...))` where both find a project, and the test checks that.

### `sysand.env.sources` export (`env/__init__.py`)

`sysand/env/_sources.py` has existed since the "only dependencies" mode was added to `sources`. It lists the sources of an installed project by IRI, whereas top-level `sysand.sources` lists them by project path. Only the top-level one was exported. This MR adds `sources` (and `projects`) to `sysand.env.__all__`; the function body is untouched.

### `EnvError` (`_errors.py`, `lib.rs::env_read_to_pyerr`)

`env_read_to_pyerr` used to wrap a failed environment read in `IOError("failed to read environment metadata: ...")`. It now raises `EnvError`, a `SysandError` subclass, with the error's own message and `wrote == False`. This affects the three existing entry points that read an environment (`env.install_path`, `env.sources`, and `sources` when `env_path` is given) as well as the new `env.projects`.

### Tests (`tests/test_basic.py`)

`_project_with_dir_usage` builds three projects with `sysand.init`: the project under test, a sibling used through a `dir` usage, and a third one installed by path under `urn:kpar:dep`. The first two are locked and synced by the shipped CLI with `--no-config --no-index`, so the environment is exactly what `sync` produces, with the extra project added afterwards because `sync` prunes anything not in the lockfile. On top of that:

- `test_env_projects_lists_installed_and_editable`: the three entries, the editable flag and relative `path` of the project's own entry, the IRIs and `lib/` path of the copied ones, and that `usages` lists the `dir` dependency's IRI.
- `test_env_sources_is_exported`: `sysand.env.sources` resolves the installed `urn:kpar:dep` project's source file and is in `__all__`.
- `test_env_projects_missing_env`: a missing environment raises `EnvError` with `wrote == False` from `projects` and from `install_path`, and the class is both a `RuntimeError` and a `SysandError`.
- `test_discover`: no project, a project found from its root and from a nested directory, agreement with `sysand.root`, a workspace found from a nested path while the workspace root itself has no project, and a malformed `.workspace.json` raising `ProjectError`.

### DIFF
--- a/bindings/py/python/sysand/_model.py
+++ b/bindings/py/python/sysand/_model.py
@@ -44,6 +44,51 @@ class UsageConstraintChange(typing.TypedDict):
     new_constraint: typing.Optional[str]
 
 
+class EnvProjectChecksumKpar(typing.TypedDict):
+    kpar_cksum: str
+    """Checksum of the KPAR the project was installed from."""
+
+
+class EnvProjectChecksumSrc(typing.TypedDict):
+    src_cksum: str
+    """Checksum of the source directory the project was installed from."""
+
+
+EnvProjectChecksum = typing.Union[EnvProjectChecksumKpar, EnvProjectChecksumSrc]
+
+
+class EnvProject(typing.TypedDict):
+    """One entry of an environment's ``env.toml``, see
+    :func:`sysand.env.projects`."""
+
+    publisher: typing.Optional[str]
+    name: str
+    version: str
+    path: str
+    """Verbatim from ``env.toml``: relative to the environment directory, or
+    to the workspace/project root when ``editable``. Not joined."""
+    identifiers: typing.List[str]
+    """IRIs of the project; the first is canonical. Empty only for
+    ``editable`` entries."""
+    usages: typing.List[str]
+    editable: bool
+    workspace: bool
+    checksum: typing.Optional[EnvProjectChecksum]
+    """Checksum of what the project was installed from, or ``None`` when
+    ``env.toml`` records none (``editable`` entries)."""
+
+
+class Discovery(typing.TypedDict):
+    """Result of :func:`sysand.discover`."""
+
+    project_root: typing.Optional[str]
+    """Canonical path of the enclosing project (the directory holding
+    ``.project.json`` or ``.meta.json``), or ``None``."""
+    workspace_root: typing.Optional[str]
+    """Canonical path of the enclosing workspace (the directory holding
+    ``.workspace.json``), or ``None``."""
+
+
 class InterchangeProjectInfo(typing.TypedDict):
     publisher: typing.Optional[str]
     name: str
--- a/bindings/py/python/sysand/_model.py
+++ b/bindings/py/python/sysand/_model.py
@@ -4,6 +4,9 @@
 from enum import Enum, auto
 import typing
 
+# IMPORTANT
+# Keep the types here in sync with Rust types from sysand-core.
+#
 # Use raw types for components, as these classes are converted
 # to/from Rust `*Raw` model type variants
 
--- a/bindings/py/src/lib.rs
+++ b/bindings/py/src/lib.rs
@@ -19,10 +19,12 @@ use sysand_core::{
         env::{EnvError, do_env_local_dir},
         init::do_init_local_file,
     },
+    discover::{discover_project, discover_workspace},
     env::{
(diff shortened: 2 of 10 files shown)


bindings/py/src/lib.rs
```diff
@@ -672,9 +723,74 @@ pub fn sysand_py(m: &Bound<'_, PyModule>) -> PyResult<()> {
...
+type EnvProjectTuple = (
+    Option<String>,
+    String,
+    String,
+    String,
+    Vec<String>,
+    Vec<String>,
+    bool,
+    bool,
+);
```
andrius-puksta-sensmetry: Return `EnvProject` structs directly, like we do for `InterchangeProjectInfo`/`InterchangeProjectMetadata`.

vytautas-astrauskas-sensmetry: Fixed in 6e12a94af5314fb8c535a8eb2108c2f79d29674a

bindings/py/python/sysand/_model.py
andrius-puksta-sensmetry, line 6:
```suggestion

# IMPORTANT
# Keep the types here in sync with Rust types from sysand-core.
#
```

andrius-puksta-sensmetry (approved):
Good to go with the suggestion applied

PR #529, opened 2026-09-08
feat(py): `AuthPolicy`, `Resolution`, typed exceptions

vytautas-astrauskas-sensmetry:
This MR gives the Python bindings the three things every index-facing call needs and that `sysand.info` was missing: a way to say which indexes to use, a way to authenticate, and exceptions a caller can catch by kind. `info` is the first user; the same objects are meant to be reused by every later function that reaches an index.

- `sysand.Resolution` mirrors the CLI's `--index` / `--default-index` / `--no-index` / `--include-std` and whether configuration files are read.
- `sysand.AuthPolicy` selects credentials: none, the CLI's own resolution from `SYSAND_CRED_*` and the credential store, or an explicit bearer token or basic pair for a URL glob.
- `sysand.info(uri, *, resolution=..., auth=...)` accepts both. The old `index_urls=` keyword still works and means the same thing.
- Eight exception classes under `SysandError`: `ResolutionError`, `NotFoundError`, `SolveError`, `AuthError`, `IndexProtocolError`, `SyncError`, plus the existing `ProjectError` and `EnvError`. `info` raises the first four that apply instead of a bare `RuntimeError`.
- The CLI crate exports `env_auth_policy()` and `standard_auth_policy(use_credential_store)`, extracted unchanged from `run_cli`, so the bindings authenticate exactly as the CLI does.

No core crate code changes.

## Why

`info` was the only binding that talked to an index, and it did so unauthenticated (the `// FIXME: Add Python support for authentication` in `do_info_py`) with the indexes named by a bare list of URLs. That has three problems for a tool driving sysand from Python:

- It cannot use a private index at all.
- It cannot read the user's `sysand` configuration, so "the indexes the CLI would use" is not expressible.
- Every failure is `RuntimeError` with a message. A caller cannot tell "no such project" from "401" from "the index is broken" without parsing text, and the text for a 401 did not say which credential was tried.

Rather than add `index_urls` and `auth` parameters to each function as it gains index access, this MR introduces the two configuration objects once, defines the exception vocabulary once, and wires `info` to them as the proof.

### `AuthPolicy` (`_auth.py`, `lib.rs::AuthSpec` / `build_auth_policy`)

A pure configuration object built through static constructors:

| constructor | meaning |
|---|---|
| `AuthPolicy.none()` | no credentials; a 401/403 raises `AuthError`. This is the default for every call. |
| `AuthPolicy.from_env(keyring=True)` | the CLI's own resolution: validated `SYSAND_CRED_<LABEL>` groups, composed with the OS credential store that `sysand auth login` writes. `keyring=False` skips the store, giving a policy that can never prompt. |
| `AuthPolicy.bearer(url_glob, token, label="python")` | one bearer token for URLs matching the glob |
| `AuthPolicy.basic(url_glob, username, password)` | one basic pair for URLs matching the glob |

Two deliberate choices:

- **Secrets never appear in `repr`.** Both constructors that take a secret print `<redacted>` in its place. `test_auth_policy_repr_hides_secrets` checks every variant.
- **Nothing happens at construction.** `from_env()` does not read the environment or open the credential store; the Rust side builds the real policy per call, inside `py.detach`. Opening the OS store can raise a keychain prompt on some platforms, so a library call only does that when the caller asked for `from_env()` with the default `keyring=True`. A malformed `SYSAND_CRED_*` group is therefore reported by the call that uses the policy, as `AuthError`, before any request is made (`test_from_env_rejects_malformed_groups`).

On the Rust side every variant becomes the CLI's own policy type (`CliAuthPolicy`, via `without_store` for the non-keyring variants), so the resolver stack is monomorphised once and `none()` is simply an empty credential map.

### `Resolution` (`_auth.py`, `lib.rs::ResolutionSpec` / `index_locations`)

Keyword-only fields `index`, `default_index`, `no_index`, `include_std`, `use_config`. `no_index` combined with either list is a `ValueError`, as the CLI rejects the equivalent flags.

`index_locations` reproduces the CLI's merge: with `use_config=True` (the default) the configuration files are loaded from the working directory and merged with the explicit lists through the same `Config::index_urls` the CLI calls, with the built-in default index as the fallback. Only the files are read. The CLI's environment overrides (`SYSAND_INDEX`, `SYSAND_DEFAULT_INDEX`, `SYSAND_CONFIG_FILE`, `SYSAND_NO_CONFIG`) are deliberately not consulted, because a library should not change behaviour based on the caller's shell. Without a `Resolution` at all, or with `no_index=True`, no index is consulted, which is what `info` did before when `index_urls` was omitted.

`include_std` is accepted and stored but not read by anything yet. It is part of the CLI's resolution options and belongs on this object; the field carries an `#[expect(dead_code)]` so the day something reads it the attribute has to go.

### `info` (`_info.py`, `lib.rs::do_info_py`)

`info(uri, *, index_urls=None, resolution=None, auth=None)`. Passing both `index_urls` and `resolution` is a `ValueError`. `index_urls` is translated to `Resolution(default_index=index_urls, use_config=False)`, which with an empty configuration yields exactly the given URLs in order, so existing callers see the same indexes as before. `test_basic.py::test_index_info`, which uses `index_urls`, is unchanged and still passes.

### Typed exceptions (`_errors.py`, `lib.rs::info_error_to_pyerr`)

All subclass `SysandError`, hence `RuntimeError`, and carry `wrote`. The hierarchy: `NotFoundError` is a `ResolutionError`; the rest are direct children.

| class | when |
|---|---|
| `NotFoundError` | no configured source knows the project |
| `AuthError` | an index answered 401 or 403, or `from_env()` found a malformed `SYSAND_CRED_*` group |
| `IndexProtocolError` | an index answered, but not with what the protocol requires (bad discovery document, `versions.json`, digest, JSON shape) |
| `ResolutionError` | any other resolution failure, including connection-level errors, which are not the index's fault |
| `SolveError` | dependency resolution found no solution; has `conflicts`, `report`, `kind` |
| `SyncError` | an installation failed part-way; has `partial` |

`info_error_to_pyerr` matches the resolver's error variants rather than walking `source()`, because the `transparent` wrappers in between collapse the source chain. The resolver's error type is known statically, so the match is exhaustive where it matters and falls back to `ResolutionError`.

`AuthError` messages carry a hint after the HTTP line, computed by `auth_hint` from the policy that was in force: which `AuthPolicy` constructor to pass when there was none, the glob that was tried for an explicit credential, or the `SYSAND_CRED_<LABEL>` variable (found, not matched, or ambiguous) for `from_env`. The hint never contains a secret; `test_info_auth_matrix` asserts the wrong token is absent from every message.

**For reviewers, the one behaviour change:** an `info` failure used to be `RuntimeError(message)`. It is now one of the typed classes above with the same message (plus the hint for `AuthError`). All of them are `RuntimeError` subclasses, so `except RuntimeError` continues to work. A malformed IRI is still `ValueError`.

`SolveError` and `SyncError` are defined here but nothing raises them yet. They are in this MR so the exception vocabulary is complete in one place and later functions do not each add a class; their attributes (`conflicts`, `report`, `kind`, `partial`) are fixed now so callers can be written against them.

### CLI crate (`sysand/src/lib.rs`)

The block of `run_cli` that built the credential policy is extracted into two public functions:

- `env_auth_policy()`: the eager part, every validated `SYSAND_CRED_*` group composed into one `StandardHTTPAuthentication`.
- `standard_auth_policy(use_credential_store)`: the above composed with the lazily opened OS credential store; an unavailable store is a warning, as before, and `false` skips the store entirely.

`run_cli` now calls `standard_auth_policy(true)`. Its behaviour is unchanged; the body moved and gained the `use_credential_store` switch that `AuthPolicy.from_env(keyring=False)` needs.

### Tests (`tests/test_auth.py`)

Seven tests against the mock index:

- `repr` hides secrets for every constructor; `Resolution` defaults and its `ValueError`/`TypeError` cases.
- The hierarchy: each class is a `SysandError` and `RuntimeError`, has `wrote == False` by default, `NotFoundError` is a `ResolutionError`, and `SolveError`/`SyncError` carry their attributes.
- `info` through `Resolution` and through `index_urls` agree; passing both raises.
- An absent project raises `NotFoundError` naming the IRI.
- The auth matrix with the index requiring a bearer token: no policy, `none()`, a wrong bearer, a basic pair (all `AuthError`, with the glob in the message and the secret absent); the right bearer succeeds and the request log shows the unauthenticated first attempt followed by the retry; `from_env()` with and without the store succeeds from `SYSAND_CRED_TEST*`; a wrong env token names the variable, not the token.
- A `SYSAND_CRED_TEST` glob with no credential raises `AuthError` from the call, before any request.

The suite's autouse fixture already clears `SYSAND_*` and points the credential store at the debug-only absent seam, so `from_env()` with `keyring=True` never touches a developer's keyring here.

### DIFF
--- a/bindings/py/src/lib.rs
+++ b/bindings/py/src/lib.rs
@@ -353,7 +353,8 @@ fn auth_hint(auth: &AuthSpec, policy: &CliAuthPolicy, url: &str) -> String {
         "bearer" | "basic" => {
             let glob = auth.url_glob.as_deref().unwrap_or("");
             format!(
-                "the {} credential for URL glob `{glob}` was rejected by, or does not match, `{url}`",
+                "the {} credential for URL glob `{glob}`\n\
+                was rejected by, or does not match, `{url}`",
                 auth.kind
             )
         }
--- a/bindings/py/src/lib.rs
+++ b/bindings/py/src/lib.rs
@@ -11,17 +11,21 @@ use pyo3::{
     types::PyAny,
 };
 use semver::{Version, VersionReq};
+use sysand::{CliAuthPolicy, DEFAULT_INDEX_URL, standard_auth_policy};
 use sysand_core::{
     add::do_add_guess,
-    auth::Unauthenticated,
+    auth::{GlobMapResult, StandardHTTPAuthenticationBuilder},
     build::{KParBuildError, KparCompressionMethod, do_build_kpar},
     commands::{
         env::{EnvError, do_env_local_dir},
         init::do_init_local_file,
     },
+    config::{Config, local_fs::load_configs},
     discover::{discover_project, discover_workspace},
     env::{
         DEFAULT_ENV_NAME, ReadEnvironment as _, WriteEnvironment as _,
+        discovery::DiscoveryError,
+        index::{HttpFetchError, IndexEnvironmentError},
         local_directory::{
             LocalDirectoryEnvironment, LocalReadError, LocalWriteError,
             metadata::{EnvMetadataError, EnvProject, EnvProjectChecksum},
@@ -31,7 +35,7 @@ use sysand_core::{
     exclude::do_exclude,
     include::do_include,
     index_location::IndexLocation,
-    info::{InfoProjectError, do_info, do_info_project},
+    info::{InfoError, InfoProjectError, do_info, do_info_project},
     init::InitError,
     model::{
         InterchangeProjectChecksumRaw, InterchangeProjectInfoRaw, InterchangeProjectMetadataRaw,
@@ -44,7 +48,12 @@ use sysand_core::{
         utils::wrapfs,
     },
     remove::do_remove_guess,
-    resolve::{net_utils::create_reqwest_client, standard::standard_resolver},
+    resolve::{
+        ResolveRead,
+        combined::CombinedResolverError,
+        net_utils::create_reqwest_client,
+        standard::{StandardResolver, standard_resolver},
+    },
     root::do_root,
     sources::{Dependencies, do_sources_local_src_project_no_deps, resolve_dependencies},
     symbols::Language,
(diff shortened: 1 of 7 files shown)


bindings/py/src/lib.rs
```diff
@@ -233,32 +434,17 @@ fn do_info_py(
...
-        let combined_resolver = standard_resolver(
-            None,
-            Some(client),
-            index_url,
-            runtime,
-            // FIXME: Add Python support for authentication
-            Arc::new(Unauthenticated {}),
-        )
-        .map_err(|err| PyValueError::new_err(format_err(err)))?;
+        // Without a `Resolution` no index is consulted, as before.
```
```suggestion
        // Without a `Resolution` no index is consulted
```

andrius-puksta-sensmetry (approved):
Good to go with the nits fixed

bindings/py/src/lib.rs
```diff
@@ -213,18 +222,210 @@ fn do_info_py_path(
...
+
+type StandardResolverError = <StandardResolver<CliAuthPolicy> as ResolveRead>::Error;
+
+/// What to tell the user about an HTTP 401/403 from `url`, given the
+/// policy that was in force. Never includes a secret.
+fn auth_hint(auth: &AuthSpec, policy: &CliAuthPolicy, url: &str) -> String {
+    match auth.kind.as_str() {
+        "none" => format!(
+            "no credentials were configured for `{url}`; pass `auth=AuthPolicy.bearer(...)`, \
+             `AuthPolicy.basic(...)` or `AuthPolicy.from_env()`"
```
```suggestion
            "no credentials were configured for `{url}`;\n\
            pass `auth=AuthPolicy.bearer(...)`, `AuthPolicy.basic(...)` or `AuthPolicy.from_env()`"
```

bindings/py/src/lib.rs
```diff
@@ -213,18 +222,210 @@ fn do_info_py_path(
...
+        ),
+        "bearer" | "basic" => {
+            let glob = auth.url_glob.as_deref().unwrap_or("");
+            format!(
+                "the {} credential for URL glob `{glob}` was rejected by, or does not match, `{url}`",
```
```suggestion
                "the {} credential for URL glob `{glob}`\n\
                was rejected by, or does not match, `{url}`",
```

bindings/py/src/lib.rs
```diff
@@ -213,18 +222,210 @@ fn do_info_py_path(
...
+                auth.kind
+            )
+        }
+        _ => match policy.env_policy().publish_bearer_auth_map() {
+            Ok(map) => match map.lookup(url) {
+                GlobMapResult::Found(entry) => format!(
+                    "the bearer token in `SYSAND_CRED_{label}_BEARER_TOKEN` (URL glob \
+                     `SYSAND_CRED_{label}`) was rejected by `{url}`",
```
```suggestion
                    "the bearer token in `SYSAND_CRED_{label}_BEARER_TOKEN` (URL glob\n\
                     `SYSAND_CRED_{label}`) was rejected by `{url}`",
```

bindings/py/src/lib.rs
```diff
@@ -213,18 +222,210 @@ fn do_info_py_path(
...
+                    label = entry.label
+                ),
+                GlobMapResult::NotFound => {
+                    format!("no `SYSAND_CRED_*` bearer token matches `{url}`")
+                }
+                GlobMapResult::Ambiguous(entries) => format!(
+                    "several `SYSAND_CRED_*` URL globs match `{url}`: {}",
```
```suggestion
                    "several `SYSAND_CRED_*` URL globs match `{url}`:\n{}",
```

PR #531, opened 2026-09-09
feat(py): `versions()` - authenticated version listing

vytautas-astrauskas-sensmetry:
# feat(py): `versions()`, authenticated version listing

This MR adds `sysand.versions(iri, *, resolution=None, auth=None)`, which lists every version a resolver can offer for a project, highest first. It is the sibling of `sysand.info`: the same resolution over the same indexes with the same `Resolution` and `AuthPolicy` objects, but every candidate's version is collected instead of keeping the best one.

- New core function `sysand_core::versions::do_versions` returning a `VersionListing { iri, versions, ignored }`. Nothing in the CLI calls it yet; the Python binding is its first user.
- `sysand.versions` returns a `VersionListing` typed dict with `versions` as strings, highest first, and `ignored` for version strings that are not valid semver. Errors are the typed classes `info` already raises: `NotFoundError`, `AuthError`, `IndexProtocolError`, `ResolutionError`.
- The combined resolver no longer probes remote and index candidates for their checksum when there is no locally cached project to match against. Over an index, a listing costs one `versions.json` request instead of two more requests per version.
- A stale doc comment in `core/src/project/index_entry.rs` is corrected. Code there is unchanged.
- `do_info_py` and the new `do_versions_py` share a `standard_resolver_for` helper; the body of `do_info_py` moved into it unchanged.

## Why

A tool driving sysand from Python has to answer "which versions of this project exist?" before it can decide whether a constraint change is even possible. `info` cannot answer that: it resolves to one project, the highest version that matches, and a caller that wants the full list would have to re-implement the index protocol. The CLI has no command for it either.

The list has to come through the resolver rather than from a direct `versions.json` fetch so that it sees exactly what `lock` will see: the same index merge, the same authentication, the same treatment of yanked and removed entries, and the same non-index sources when a `Resolution` names them.

### `do_versions` (`core/src/commands/versions.rs`)

Resolves `iri` exactly as `do_info` does, then iterates every candidate and collects `project.version()` instead of picking the best one. Over an index each candidate answers `version()` from the entry advertised in `versions.json`, so no per-version fetch is needed.

Semantics, each covered by a unit test in `versions_tests.rs`:

- `versions` is sorted highest first with duplicates removed (the same version offered by two sources is listed once).
- Strings that are not valid semver go to `ignored`, in encounter order. An index cannot publish such entries (ingest validation rejects them), but a `dir` or `src` source can, and dropping them silently would hide a misconfigured project.
- A candidate that fails to load or declares no version is skipped with a log line, as `do_info` skips them.
- If every candidate was skipped the result is `InfoError::NoSemanticVersionsFound` with an empty list. If candidates were found but none had a semver version, the result is `Ok` with an empty `versions` and a full `ignored`, so the caller can report what it saw.
- `NotFound`, `UnsupportedUsage` and `NoResolve` map to the same `InfoError` variants as in `do_info`, so the Python error mapping written for `info` applies without change.

The error type is `InfoError<R::Error>`, shared with `do_info`, rather than a new one: the failure modes are identical and the Python side already maps every variant.

### Combined resolver: skip the checksum probe when nothing can match (`core/src/resolve/combined.rs`)

`CombinedIterator` matches each remote or index candidate against locally cached copies by canonical checksum. For an index candidate that checksum is computed from `.project.json` and `.meta.json`, which means two fetches per candidate, for every version in the listing. When the local set is empty the result of the lookup is always "no match", so the probe is pure cost.

The two duplicated match blocks (one for remote, one for index candidates) are replaced by `take_cached_local`, which returns `None` up front when `locals` is empty and otherwise does what the two blocks did. Two new tests in `combined_tests.rs` use a `CountingProject` wrapper to assert that candidates are not read when there is nothing local to match, and are read when a local copy is waiting.

**For reviewers:** the results are identical by construction: with an empty local set the old code always ended in `shift_remove` on an empty map. What changes is the number of requests, and it differs by candidate kind:

- Index candidates answer `version()` from `versions.json`, so the saving only shows up for callers that need nothing but `version()` or `usage()`. `do_info` is not one of them: it calls `get_project()` on every candidate to pick the best, so `sysand info <iri>` and `sysand.info` over an index make the same requests as before (checked against the mock index, three versions published, six `.project.json`/`.meta.json` fetches before and after).
- Remote candidates (a project given by URL) have no cached listing to answer from. Their checksum probe re-resolved the URL: one more `info/refs?service=git-upload-pack` probe, one more `GET` of the URL as a kpar, and one more `GET` each of `.project.json` and `.meta.json`. With nothing local to match, that round is gone for every URL-based project in `info`, `lock`, `sync` and `env install`. The CLI tests in `sysand/tests/cli_env.rs`, `cli_info.rs` and `cli_lock.rs` pin these counts and are lowered by one per URL-based candidate; most of the old values carried a `TODO: Reduce this to 1` comment, and the git probe ones are now at that target. Behaviour and output are unchanged; only the mock expectations move.

The `checksum_canonical_hex failed; skipping local-cache match` debug line is no longer logged when the local set is empty, since the probe that could fail is not made.

### `index_entry.rs` doc fix

The module and field docs said `checksum_canonical_hex_async` returns the advertised digest without I/O until the archive is verified. It does not, and did not before this MR: the canonical project digest is computed from the fetched info and metadata, which is exactly why the probe above was expensive. The advertised `kpar_digest` is returned by `checksum_canonical_variant_async`, whose docs were already correct. Only the comments change.

### `sysand.versions` (`_versions.py`, `_model.py`, `lib.rs::do_versions_py`)

```python
sysand.versions("pkg:sysand/acme/lib", resolution=sysand.Resolution(...), auth=sysand.AuthPolicy.bearer(...))
# -> {"iri": "pkg:sysand/acme/lib", "versions": ["1.2.0", "1.1.0"], "ignored": []}
```

`VersionListing` is a typed dict with `iri`, `versions` (strings, highest first) and `ignored`. The Rust side returns a plain tuple of strings and the Python side builds the dict, as the other bindings do.

**For reviewers, one deliberate difference from `info`:** when `resolution` is omitted, `versions` uses `Resolution()`, meaning the CLI's configuration files plus the default index, whereas `info` with no `Resolution` consults no index at all. `info` kept that behaviour for compatibility with its original `index_urls=None` meaning; a version listing with no index to ask is useless, so `versions` defaults to what the CLI would use. There is no `index_urls` keyword on `versions` because it is new and has no callers to keep working.

Error mapping reuses `info_error_to_pyerr`: `NotFoundError` when no source knows the IRI (including `Resolution(no_index=True)` with nothing local), `AuthError` with the same credential hint on 401/403, `IndexProtocolError` when `versions.json` is malformed or violates the protocol, `ResolutionError` for everything else, including the "candidates found but none has a version" case.

`do_info_py` is refactored to share `standard_resolver_for` (HTTP client, current-thread runtime, index merge, auth policy) and `parse_iri` with `do_versions_py`. Its behaviour is unchanged and `test_auth.py` still passes untouched.

### Mock index: `override(path, body)` (`tests/mockindex.py`)

Serves an arbitrary body for a path regardless of what was published. Used to feed the resolver a `versions.json` that is valid JSON but breaks the protocol, and one that is not JSON at all.

### Tests (`tests/test_versions.py`)

Five tests against the mock index. Every call passes an explicit `resolution=` because the default would touch the public index.

- Three published versions come back highest first, and the request log shows exactly one `versions.json` request and no `.project.json`, `.meta.json` or `project.kpar` requests. This is the end-to-end check of the combined-resolver change.
- A `pkg:sysand/...` IRI works through the PURL path layout.
- An absent IRI raises `NotFoundError` with `wrote == False`, as does `Resolution(no_index=True)`; the log shows the index was asked only for the absent one.
- With the index requiring a bearer token: no policy raises `AuthError`; `AuthPolicy.bearer` and `AuthPolicy.from_env(keyring=False)` from `SYSAND_CRED_TEST*` both succeed.
- A `versions.json` in ascending order (valid JSON, invalid protocol) and a non-JSON body both raise `IndexProtocolError` with `wrote == False`.

### DIFF
--- a/core/src/commands/versions.rs
+++ b/core/src/commands/versions.rs
@@ -1,0 +1,121 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+// SPDX-FileCopyrightText: © 2026 Sysand contributors <opensource@sensmetry.com>
+
+//! List every version a resolver can offer for a project.
+
+use fluent_uri::Iri;
+use semver::Version;
+
+use crate::{
+    info::InfoError,
+    project::ProjectRead as _,
+    resolve::{ResolutionInfo, ResolutionOutcome, ResolveRead},
+    utils::format_err,
+};
+
+/// The versions a resolver offers for one project.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct VersionListing {
+    /// The IRI that was looked up.
+    pub iri: String,
+    /// Distinct semver-valid versions, highest first. Over an index these
+    /// are the *available* versions: yanked and removed entries are never
+    /// enumerated.
+    pub versions: Vec<Version>,
+    /// Version strings that are not valid semver, in encounter order. Kept
+    /// for diagnostics rather than dropped silently; an index cannot publish
+    /// such entries, but other sources can.
+    pub ignored: Vec<String>,
+}
+
+/// List the versions of `iri` known to `resolver`.
+///
+/// The sibling of [`crate::info::do_info`]: the same resolution, but every
+/// candidate's version is collected instead of keeping the best one. Over an
+/// index this costs the per-project `versions.json` request only — each
+/// candidate answers `version()` from the advertised entry.
+///
+/// Candidates that fail to load, or report no version, are skipped with a
+/// log line, as `do_info` skips them. If every candidate was skipped the
+/// result is [`InfoError::NoSemanticVersionsFound`] with an empty list; if
+/// only non-semver versions were found the listing is `Ok` with an empty
+/// `versions` and a full `ignored`, so the caller can report them.
+#[expect(clippy::result_large_err)]
+pub fn do_versions<R: ResolveRead>(
+    iri: &Iri<String>,
+    resolver: &R,
+) -> Result<VersionListing, InfoError<R::Error>> {
+    let resolve = ResolutionInfo::iri(iri.to_owned());
+    let outcome = resolver.resolve_read(&resolve)?;
+
+    let resolved = match outcome {
+        ResolutionOutcome::Resolved(resolved) => resolved,
+        ResolutionOutcome::UnsupportedUsageType { reason } => {
+            return Err(InfoError::UnsupportedUsage {
+                usage: resolve,
+                reason,
+            });
+        }
+        ResolutionOutcome::NotFound { reason } => {
+            return Err(InfoError::NotFound {
+                usage: resolve,
+                reason,
+            });
+        }
+        ResolutionOutcome::Unresolvable { reason } => {
+            return Err(InfoError::NoResolve {
+                usage: resolve,
+                reason,
+            });
+        }
+    };
+
+    let mut versions: Vec<Version> = Vec::new();
+    let mut ignored: Vec<String> = Vec::new();
+    let mut seen_any = false;
+
+    for candidate in resolved {
+        let project = match candidate {
+            Ok(project) => project,
+            Err(e) => {
+                // As in `do_info`: the candidate list is every possibility,
+                // and some are expected not to work.
+                log::debug!("skipping candidate project: {}", format_err(e));
+                continue;
+            }
+        };
+        match project.version() {
+            Ok(Some(version)) => {
+                seen_any = true;
+                match Version::parse(&version) {
+                    Ok(parsed) => versions.push(parsed),
+                    Err(_) => ignored.push(version),
+                }
+            }
+            Ok(None) => {
+                log::warn!("ignoring a candidate of `{iri}` that declares no version");
+            }
+            Err(err) => {
+                log::warn!("ignoring a project because: {}", format_err(err));
+            }
+        }
+    }
+
+    if !seen_any {
+        return Err(InfoError::NoSemanticVersionsFound(Vec::new()));
+    }
+
+    // Reverse sort
+    versions.sort_unstable_by(|a, b| b.cmp(a));
+    versions.dedup();
+
+    Ok(VersionListing {
+        iri: iri.to_string(),
+        versions,
+        ignored,
+    })
+}
+
+#[cfg(test)]
+#[path = "./versions_tests.rs"]
+mod tests;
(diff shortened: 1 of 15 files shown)


core/src/resolve/combined_tests.rs
```diff
@@ -392,3 +392,131 @@ fn no_semantic_versions_error() {
 
     assert_matches!(info_meta, Err(InfoError::NoSemanticVersionsFound(_)));
 }
+
+/// An in-memory project that counts how often its info/meta are read, so a
+/// test can assert that a candidate was *not* inspected.
+#[derive(Clone, Debug)]
+struct CountingProject {
+    inner: InMemoryProject,
+    reads: std::rc::Rc<std::cell::Cell<usize>>,
```
andrius-puksta-sensmetry: Import stuff (many places in this file). Fully qualified names should only be used when ambiguous.

core/src/resolve/combined_tests.rs
```diff
@@ -392,3 +392,131 @@ fn no_semantic_versions_error() {
...
+    let outcome = resolver
+        .resolve_read(&ResolutionInfo::iri(Iri::parse(uri).unwrap().into()))
+        .unwrap();
+    let ResolutionOutcome::Resolved(candidates) = outcome else {
+        panic!("expected the IRI to resolve");
+    };
+    let _candidates: Vec<_> = candidates.collect();
+
+    assert!(
+        reads.get() > 0,
```
andrius-puksta-sensmetry: Precision, i.e. `assert_eq!` would be good here.

core/src/commands/versions.rs
```diff
@@ -0,0 +1,121 @@
...
+        }
+    }
+
+    if !seen_any {
+        return Err(InfoError::NoSemanticVersionsFound(Vec::new()));
+    }
+
+    versions.sort_unstable();
+    versions.dedup();
+    versions.reverse();
```
andrius-puksta-sensmetry:
```suggestion
    // Reverse sort
    versions.sort_unstable_by(|a, b| b.cmp(a));
    versions.dedup();
```

andrius-puksta-sensmetry (approved):
Good to go when nits are fixed

core/src/resolve/combined_tests.rs
vytautas-astrauskas-sensmetry: Updated.

core/src/resolve/combined_tests.rs
vytautas-astrauskas-sensmetry: Updated.

PR #532, opened 2026-09-09
feat(py): `lock()` with the resolved set and structured solve failures

vytautas-astrauskas-sensmetry:
This MR adds `sysand.lock(path, *, resolution=None, auth=None, provided=None, write=True)`. It runs the same resolution as `sysand lock`, returns the resolved set together with the canonical lockfile text, and can do so without writing anything. When the solve fails it raises `SolveError` carrying a machine-readable list of the constraints that clashed, in addition to the CLI's report text.

- core (solver): `SolverError` now keeps the candidates the solver had resolved when it failed and offers `conflicts()`, `kind()` and `resolution_error()`. `Display` is unchanged; it remains the human report. `InternalSolverError::VersionNotAvailable` becomes a struct variant holding the resource, constraint and found versions; its message is reproduced verbatim.
- sysand (CLI crate): `command_lock` is split into `resolve_lock` (solve, no write, with caller-provided projects) plus the write. The new `CliResolver` and `CliLockError` type aliases name the resolver stack the CLI builds and the error it fails with, so the bindings can match on them. CLI behaviour is unchanged.
- py: `sysand.lock` returns a `LockResult` with `projects` (one `LockedProject` per lockfile entry) and `text`. Failures raise `SolveError` with `.conflicts`, `.report` and `.kind`, or the typed resolution errors `info` already raises when the solver hit a transport or authentication problem.
- tests: solver conflict shapes in core, `sysand.lock` against the mock index, and three new end-to-end scenarios in `test_migration_scenarios.py`.

## Why

A tool that changes a usage constraint and re-locks has to be able to show the user what will move before anything is written, and when the solve fails it has to say which pin excludes the wanted version. Neither is possible through the CLI: `sysand lock` writes on success and prints a pubgrub derivation tree on failure. That tree is rendered as prose. Recovering "dependent X pins library to `^0.10.1`" from it means parsing English.

The solver already knows every participant of a failure. This MR exposes that knowledge as data, keeps the human report byte-for-byte as it was, and gives Python the same solve the CLI runs, with a `write=False` switch.

### Structured solver failures (`core/src/solve/pubgrub.rs`)

`SolverError` gains a `candidates` map, filled by `solve()` on failure from the candidates it had resolved so far. pubgrub speaks in candidate indices (`DiscreteHashSet`), and this map is what turns an index back into a version and a dependent's usage back into its constraint.

On top of it:

- `Display` is untouched: it is still the human report, and `to_string()` is how the bindings obtain it. The CLI prints exactly what it printed before (asserted verbatim in `pubgrub_tests.rs` for the no-matching-version case).
- `kind()` is `"no_solution"` (the constraints contradict each other), `"retrieval"` (a project or version could not be obtained) or `"choosing_version"`.
- `resolution_error()` returns the resolver's own error when a retrieval failure was caused by resolution rather than by the solve. Callers use it to tell a 401 from a genuine conflict.
- `conflicts()` flattens the failure into `SolveConflict` values, in encounter order without duplicates:
  - `Constraint { iri, constraint, required_by }`: `iri` is constrained to `constraint` by `required_by`, `None` when the requirement is the root's own.
  - `NoVersions { iri, constraint, found, required_by }`: `iri` exists but none of `found` satisfies `constraint`.
  - `NotFound { iri, reason }`: `iri` could not be retrieved at all.

  For `NoSolution` the derivation tree is walked and each `External` node becomes one conflict. For `ErrorRetrievingDependencies` the single cause becomes a `NoVersions` or `NotFound`. Transport failures and version-choice failures produce an empty list; the former are what `resolution_error()` is for.

**For reviewers, the one subtle part:** pubgrub interns packages by identifier. When two dependents use the same library, the `dependency` package in a `FromDependencyOf` edge carries whichever usage the solver saw first, so reading the constraint off that package would attribute the wrong pin. `conflicts()` therefore recovers the constraint from the dependent's side: the root request's usages for `Requested`, or the usages of the dependent's candidates selected by `dependent_set` for `Remote`. When a dependent lists the same resource more than once, `select_usage` picks the usage whose constraint selects exactly the candidates in `dependency_set`, falling back to the first. `conflicts_name_the_dependents_that_pin_transitively` is the test for this.

`InternalSolverError::VersionNotAvailable(String)` becomes `VersionNotAvailable { resource, constraint, found }`. The `#[error]` attribute renders the same four-line message the string used to hold, with `format_found_versions` producing the `` `v1`, `v2` `` list; `no_matching_version_is_a_retrieval_failure_with_the_found_versions` asserts the exact text.

### `resolve_lock` (`sysand/src/commands/lock.rs`)

`command_lock` did three things: build the provided set, solve, write. The first two move into `resolve_lock`, which additionally takes `extra_provided: ProvidedProjects` merged into the standard-library set (or into an empty set with `include_std`). `command_lock` calls it with an empty `extra_provided` and writes, so the CLI path is the same code as before.

`CliResolver<Policy>` spells out the resolver `create_resolver` builds (overrides, then provided projects, then the standard file/env/index stack), and `CliLockError<Policy>` is the `LockProjectError` instantiation that `resolve_lock` fails with, reachable from the returned `anyhow::Error` by `downcast_ref`. Both aliases exist only so the bindings can name these types; nothing in the CLI uses them.

### `sysand.lock` (`_lock.py`, `_model.py`, `lib.rs::do_lock_py`)

```python
result = sysand.lock(project_dir, resolution=sysand.Resolution(...), write=False)
result["projects"]   # [{"publisher", "name", "version", "identifiers", "exports", "usages", "sources"}, ...]
result["text"]       # the canonical sysand-lock.toml, byte-exact
```

- `path` is where discovery starts, like the CLI's working directory: the enclosing project and workspace are found by walking up, the environment root is the workspace root or else the project root, and the lockfile belongs to the project root. Calling from a nested directory writes the lockfile at the project root, not in the nested directory. This mirrors the steps `run_cli` performs before `command_lock`, so the resolution matches the CLI's.
- `write=True` (the default, as in the CLI) writes `sysand-lock.toml` at the project root, and `text` is exactly what was written. `write=False` writes nothing; `text` is what would have been written. The intended sequence is a dry run, a diff shown to the user, then a second call with `write=True`.
- `resolution` defaults to `Resolution()`, the CLI's configuration plus default index, as `versions` does. `auth` defaults to `AuthPolicy.none()`.
- `provided` is a sequence of `ProvidedProject` dicts (`iri`, `info`, `meta`). They are treated as already present, the way the standard libraries are: the solver sees them, they are never fetched, and they appear in `projects` with empty `sources`.

`LockedProject` fields are the lockfile's: `usages` are the lockfile usage strings (this is what names who pins what), `sources` are the lockfile renderings of each source, kept opaque and meant for diagnostics.

### Error mapping (`lib.rs`)

The solve runs with the GIL released. Typed exceptions with keyword arguments cannot be built without it, so failures are collected in a `Failure` enum and turned into exceptions once the thread is attached again:

- A solver failure becomes `SolveError(message, conflicts=..., report=..., kind=...)`. Each `SolveConflict` becomes a dict with a `"kind"` key and the variant's fields.
- If the solver failed because the resolver failed (`resolution_error()` is `Some`), the error is classified the way `info` classifies it: `AuthError` with the credential hint on 401/403, `IndexProtocolError` for malformed index answers, `ResolutionError` otherwise. `info_error_to_pyerr` is refactored to share this classification through a new `combined_error_to_pyerr`; `info`'s behaviour is unchanged.
- Not inside a project, unreadable configuration, or any other `resolve_lock` failure: `ProjectError` with `wrote == False`.
- The lockfile write itself failing after a successful solve: `ProjectError` with `wrote == True`.

`SolveError` itself already existed in `_errors.py` with these keyword arguments; this MR is the first to raise it. Its docstring is extended to list `"choosing_version"` among the kinds, since the Rust side can produce it.

### Tests

Core (`pubgrub_tests.rs`, `lock_tests.rs`):

- Two root pins that contradict each other are both reported with `required_by == None`.
- Two dependents pinning the same library to different exact versions are each reported with their own constraint, and no `NotFound` appears.
- A constraint no version satisfies is a `"retrieval"` failure with one `NoVersions` conflict listing the found versions, and the CLI's message is unchanged, asserted verbatim.
- An unknown project is a `NotFound` conflict with `resolution_error()` empty.
- Through `do_lock_projects` with an in-memory resolver: a root wanting `library >=0.11.0, <0.12.0` while a dependent it also uses pins `library "0.10.1"` (caret). Both the dependent's `^0.10.1` and the root's own constraint are named.

Python (`tests/test_lock.py`, nine tests against the mock index, every call with an explicit `resolution=`):

- `write=False` leaves no lockfile and returns the resolved set, including the dependency's publisher, identifiers, a source carrying the archive digest, and the root's usages.
- `write=True` writes exactly `text`, and a dry run and a real run produce the same text.
- Locking from a nested directory writes the lockfile at the project root.
- A conflict names the dependent: `kind == "no_solution"`, `report` is contained in `str(error)`, both the dependent's `^0.10.1` and the root's `>=0.11.0, <0.12.0` are in `conflicts`, no lockfile is written.
- No matching version: `kind == "retrieval"`, exactly one `NoVersions` conflict with `found == ["0.10.3"]`.
- An absent project is a `NotFound` conflict.
- Outside a project: `ProjectError` with `wrote == False`.
- A `provided` project resolves at the provided version, has no sources, and the index is never asked for it.
- With the index requiring a bearer token: `AuthError` with `wrote == False`, then success with `AuthPolicy.bearer`.

### Tests (`tests/test_migration_scenarios.py`)

Three scenarios, continuing the suite from the `baseline` fixture. Each moves the library constraint to the 0.11 line with `set_usage_constraint`, then calls `sysand.lock(..., write=False)` and checks that the failure is attributable and that nothing on disk changed (`baseline.snapshot()` before and after).

- `test_dependent_excludes_new_version`: a second project used by the root pins the library to `"0.10.1"` (caret, `<0.11.0`). The baseline resolves; the migrated manifest does not, and `conflicts` contains a `Constraint` on the library with `required_by` equal to the dependent. Lockfile, `env.toml` and installed set are untouched.
- `test_no_compatible_dependent_release`: the same shape built through a `_dependent_baseline` helper that publishes the dependent with a configurable constraint on the library, so later scenarios can reuse it.
- `test_new_version_not_published_yet`: before 0.11.0 exists, `versions` lists only 0.10.3 and `lock` fails with one `NoVersions` conflict naming the library, the target constraint and `found == ["0.10.3"]`. Once 0.11.0 is published the same dry-run call resolves to it and still writes nothing.

The module docstring is updated to mention `lock`, and two small helpers are added: `resolution(index)` builds the `Resolution` for the mock index, `project_version(projects, iri)` picks a version out of a `LockResult`.

### DIFF
--- a/bindings/py/python/sysand/_lock.py
+++ b/bindings/py/python/sysand/_lock.py
@@ -1,0 +1,63 @@
+# SPDX-License-Identifier: MIT OR Apache-2.0
+# SPDX-FileCopyrightText: © 2026 Sysand contributors <opensource@sensmetry.com>
+
+from __future__ import annotations
+
+import typing
+from pathlib import Path
+
+import sysand._sysand_core as sysand_rs  # type: ignore
+
+from ._auth import AuthPolicy, Resolution
+from ._model import LockResult, ProvidedProject
+
+
+def lock(
+    path: str | Path = ".",
+    *,
+    resolution: Resolution | None = None,
+    auth: AuthPolicy | None = None,
+    provided: typing.Sequence[ProvidedProject] | None = None,
+    write: bool = True,
+) -> LockResult:
+    """Resolve the project's dependencies into a lockfile, as ``sysand lock``
+    does, and return the resolved set.
+
+    ``path`` is where discovery starts, like the CLI's working directory:
+    the enclosing project (``.project.json``) and workspace are found by
+    walking up, and the environment root is the workspace root, else the
+    project root. The lockfile belongs to the project root.
+
+    With ``write=True`` (the default, matching the CLI) ``sysand-lock.toml``
+    is written at the project root; the returned ``text`` is exactly what
+    was written. With ``write=False`` nothing is written and ``text`` is
+    what would have been — call it first, show the user the delta, then
+    lock again with ``write=True``.
+
+    ``resolution`` defaults to :class:`Resolution` ``()`` (the CLI's
+    semantics); ``auth`` defaults to :meth:`AuthPolicy.none`. ``provided``
+    projects are treated as already present, like the standard libraries.
+
+    Raises:
+        SolveError: no compatible set of versions exists. ``conflicts`` names
+            every participant (which pin excludes what), ``report`` is the
+            CLI's own text, ``wrote`` is ``False``.
+        ProjectError: ``path`` is not inside a project, the project is
+            malformed, or (with ``wrote=True``) the lockfile could not be
+            written after a successful solve.
+        AuthError, IndexProtocolError, ResolutionError: an index could not be
+            used.
+    """
+    if resolution is None:
+        resolution = Resolution()
+    text, projects = sysand_rs.do_lock_py(
+        str(path),
+        resolution._spec(),
+        auth._spec() if auth is not None else None,
+        list(provided or ()),
+        write,
+    )
+    return LockResult(projects=projects, text=text)
+
+
+__all__ = ["lock"]
--- a/bindings/py/src/lib.rs
+++ b/bindings/py/src/lib.rs
@@ -6,12 +6,18 @@ use std::{iter, process::ExitCode, sync::Arc};
 use camino::{Utf8Path, Utf8PathBuf};
 use fluent_uri::Iri;
 use pyo3::{
+    PyTypeInfo as _,
     exceptions::{PyFileExistsError, PyFileNotFoundError, PyIOError, PyRuntimeError, PyValueError},
     prelude::*,
-    types::PyAny,
+    types::{PyAny, PyDict, PyType},
 };
 use semver::{Version, VersionReq};
-use sysand::{CliAuthPolicy, DEFAULT_INDEX_URL, standard_auth_policy};
+use sysand::{
+    CliAuthPolicy, DEFAULT_INDEX_URL,
+    cli::ResolutionOptions,
+    commands::lock::{CliLockError, resolve_lock},
+    get_env, standard_auth_policy,
+};
 use sysand_core::{
     add::do_add_guess,
     auth::{GlobMapResult, StandardHTTPAuthenticationBuilder},
@@ -19,8 +25,10 @@ use sysand_core::{
     commands::{
         env::{EnvError, do_env_local_dir},
         init::do_init_local_file,
+        lock::{DEFAULT_LOCKFILE_NAME, LockError, LockProjectError},
     },
     config::{Config, local_fs::load_configs},
+    context::ProjectContext,
     discover::{discover_project, discover_workspace},
     env::{
         DEFAULT_ENV_NAME, ReadEnvironment as _, WriteEnvironment as _,
@@ -37,6 +45,7 @@ use sysand_core::{
     index_location::IndexLocation,
     info::{InfoError, InfoProjectError, do_info, do_info_project},
     init::InitError,
+    lock::{Lock, Project as LockedProject},
     model::{
         InterchangeProjectChecksumRaw, InterchangeProjectInfoRaw, InterchangeProjectMetadataRaw,
         InterchangeProjectUsage, InterchangeProjectUsageRaw,
@@ -45,19 +54,23 @@ use sysand_core::{
         ProjectRead as _,
         local_kpar::{KparInnerPath, LocalKParProject},
         local_src::{EditInfoError, LocalSrcError, LocalSrcProject},
-        utils::wrapfs,
+        memory::InMemoryProject,
+        utils::{Identifier, wrapfs},
     },
     remove::do_remove_guess,
     resolve::{
         ResolveRead,
         combined::CombinedResolverError,
         net_utils::create_reqwest_client,
+        priority::PriorityError,
         standard::{StandardResolver, standard_resolver},
     },
     root::do_root,
+    solve::pubgrub::SolveConflict,
     sources::{Dependencies, do_sources_local_src_project_no_deps, resolve_dependencies},
     symbols::Language,
     usage::{ConstraintChange, SetConstraintError, do_set_usage_constraint_local},
+    utils::ProvidedProjects,
     utils::format_err,
     versions::do_versions,
 };
@@ -265,8 +278,6 @@ struct ResolutionSpec {
     default_index: Vec<String>,
     #[pyo3(default)]
     no_index: bool,
-    /// Read by `lock`/`sync`, which build the provided-projects set.
-    #[expect(dead_code, reason = "read only by lock and sync, not exposed yet")]
     #[pyo3(default)]
     include_std: bool,
     use_config: bool,
@@ -394,10 +405,22 @@ fn info_error_to_pyerr(
     let message = format_err(&err);
     match &err {
         InfoError::NotFound { .. } => PyNotFoundError::new_err(message),
-        InfoError::Resolution(CombinedResolverError::Index(
+        InfoError::Resolution(inner) => combined_error_to_pyerr(inner, message, auth, policy),
+        _ => PyResolutionError::new_err(message),
+    }
+}
+
+fn combined_error_to_pyerr<F, L, R>(
+    err: &CombinedResolverError<F, L, R, IndexEnvironmentError>,
+    message: String,
+    auth: &AuthSpec,
+    policy: &CliAuthPolicy,
+) -> PyErr {
+    match err {
+        CombinedResolverError::Index(
             IndexEnvironmentError::Discovery(DiscoveryError::Fetch(fetch))
             | IndexEnvironmentError::Fetch(fetch),
-        )) => match fetch {
+        ) => match fetch {
             HttpFetchError::BadHttpStatus { url, status }
                 if matches!(status.as_u16(), 401 | 403) =>
             {
@@ -407,9 +430,7 @@ fn info_error_to_pyerr(
             HttpFetchError::Request { .. } => PyResolutionError::new_err(message),
             _ => PyIndexProtocolError::new_err(message),
         },
-        InfoError::Resolution(CombinedResolverError::Index(_)) => {
-            PyIndexProtocolError::new_err(message)
-        }
+        CombinedResolverError::Index(_) => PyIndexProtocolError::new_err(message),
         _ => PyResolutionError::new_err(message),
     }
 }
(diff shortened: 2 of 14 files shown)


bindings/py/python/sysand/_lock.py
```diff
@@ -0,0 +1,85 @@
...
+            )
+            for (
+                publisher,
+                name,
+                version,
+                identifiers,
+                exports,
+                usages,
+                sources,
+            ) in projects
```
andrius-puksta-sensmetry: Use a Rust struct for this.

vytautas-astrauskas-sensmetry: Updated.

core/src/solve/pubgrub.rs
```diff
@@ -499,6 +542,237 @@ impl<R: ResolveRead + fmt::Debug + 'static> From<pubgrub::PubGrubError<ProjectSo
...
+                    | InternalSolverError::Unresolvable { usage, .. }
+                    | InternalSolverError::InvalidProject { usage, .. }
+                    | InternalSolverError::MissingVersion { usage }
+                    | InternalSolverError::MissingUsage { usage }
+                    | InternalSolverError::InvalidResolvedVersion { usage, .. }
+                    | InternalSolverError::VersionObtain { usage, .. }
+                    | InternalSolverError::UsageObtain { usage, .. } => {
+                        conflicts.push(SolveConflict::NotFound {
+                            iri: usage.id().to_string(),
+                            reason: source.to_string(),
```
andrius-puksta-sensmetry: Use `format_err`.

vytautas-astrauskas-sensmetry: Updated.

core/src/solve/pubgrub.rs
```diff
@@ -499,6 +542,237 @@ impl<R: ResolveRead + fmt::Debug + 'static> From<pubgrub::PubGrubError<ProjectSo
...
+            } => candidates
+                .iter()
+                .enumerate()
+                .all(|(i, c)| constraint.matches(&c.version) == dependency_set.contains(&i)),
+            _ => false,
+        };
+        if selects_set(first) {
+            return Some(first);
+        }
+        Some(usages.find(|u| selects_set(u)).unwrap_or(first))
```
andrius-puksta-sensmetry, line 812:
```suggestion
    fn select_usage<'a>(
        &self,
        mut usages: impl Iterator<Item = &'a CoalescingUsage>,
        dependency: &Identifier,
        dependency_set: &DiscreteHashSet,
    ) -> Option<&'a CoalescingUsage> {
        let candidates: Vec<&CandidateSnapshot> = self
            .candidates
            .get(dependency)
            .into_iter()
            .flatten()
            .collect();
        let selects_set = |usage: &CoalescingUsage| match usage.usage().usage() {
            InterchangeProjectUsage::Resource {
                version_constraint: Some(constraint),
                ..
            } => candidates
                .iter()
                .enumerate()
                .all(|(i, c)| constraint.matches(&c.version) == dependency_set.contains(&i)),
            _ => false,
        };
        Some(usages.find(selects_set).unwrap_or(first))
```

core/src/solve/pubgrub.rs
```diff
@@ -499,6 +542,237 @@ impl<R: ResolveRead + fmt::Debug + 'static> From<pubgrub::PubGrubError<ProjectSo
...
+            .flatten()
+            .collect();
+        let selects_set = |usage: &CoalescingUsage| match usage.usage().usage() {
+            InterchangeProjectUsage::Resource {
+                version_constraint: Some(constraint),
+                ..
+            } => candidates
+                .iter()
+                .enumerate()
+                .all(|(i, c)| constraint.matches(&c.version) == dependency_set.contains(&i)),
```
andrius-puksta-sensmetry: How is `i` related to `dependency_set`?

core/src/solve/pubgrub.rs
```diff
@@ -499,6 +542,237 @@ impl<R: ResolveRead + fmt::Debug + 'static> From<pubgrub::PubGrubError<ProjectSo
...
+        Some(usages.find(|u| selects_set(u)).unwrap_or(first))
+    }
+
+    fn versions_of(&self, id: &Identifier, set: &DiscreteHashSet) -> Vec<String> {
+        let mut found: Vec<String> = self
+            .candidates_in(id, set)
+            .into_iter()
+            .map(|candidate| candidate.version.to_string())
+            .collect();
+        found.dedup();
```
andrius-puksta-sensmetry: `dedup` only removes directly adjacent duplicates. Candidates have to be sorted beforehand to group duplicates.
String sorting is not correct in semver sense, but works for this purpose.
```suggestion
        found.sort_unstable()
        found.dedup();
```

core/src/solve/pubgrub.rs
```diff
@@ -499,6 +542,237 @@ impl<R: ResolveRead + fmt::Debug + 'static> From<pubgrub::PubGrubError<ProjectSo
...
+                                        dependency.id(),
+                                        dependency_set,
+                                    ),
+                                ),
+                            };
+                            SolveConflict::Constraint {
+                                iri,
+                                constraint: constraint_of(constraint.unwrap_or(dependency)),
+                                required_by,
+                            }
```
andrius-puksta-sensmetry, line 748: I don't like this hacky workaround, but fixing it requires changing how we represent version constraints in the solver. For now just add clarification and a `FIXME`:
```suggestion
                        // pubgrub interns packages by identifier, so the
                        // `dependency` package may carry another party's
                        // usage, so only the identifier of it is correct to use here.
                        // Because `dependency_set` is currently `DiscreteHashSet` which
                        // does not include version constraints in a user-readable form,
                        // version constraints are instead extracted from the usage list
                        // of `dependent`.
                        // FIXME: make `dependency_set` usable directly
                        id_of(dependency).map(|dependency| {
                            let iri = dependency.id().to_string();
                            let (required_by, constraint) = match dependent {
                                DependencyIdentifier::Requested(usages) => (
                                    None,
                                    self.select_usage(
                                        usages.iter().filter(|u| u.id() == dependency.id()),
                                        dependency.id(),
                                        dependency_set,
                                    ),
                                ),
                                DependencyIdentifier::Remote(usage) => (
                                    Some(usage.id().to_string()),
                                    self.select_usage(
                                        self.candidates_in(usage.id(), dependent_set)
                                            .into_iter()
                                            .flat_map(|c| c.usages.iter())
                                            .filter(|u| u.id() == dependency.id()),
                                        dependency.id(),
                                        dependency_set,
                                    ),
                                ),
                            };
                            SolveConflict::Constraint {
                                iri,
                                constraint: constraint_of(constraint.unwrap_or(dependency)),
                                required_by,
                            }
```

andrius-puksta-sensmetry (commented):
Haven't finished reviewing yet.

bindings/py/src/lib.rs
```diff
@@ -394,10 +405,25 @@ fn info_error_to_pyerr(
...
         InfoError::NotFound { .. } => PyNotFoundError::new_err(message),
-        InfoError::Resolution(CombinedResolverError::Index(
+        InfoError::Resolution(inner) => combined_error_to_pyerr(inner, message, auth, policy),
+        _ => PyResolutionError::new_err(message),
+    }
+}
+
+/// Typed exception for a failure of the standard resolver stack:
+/// HTTP 401/403 → `AuthError` (with a hint), connection failures →
+/// `ResolutionError`, malformed index answers → `IndexProtocolError`.
```
```suggestion
```
Don't restate what is obvious in the code.

core/src/solve/pubgrub.rs
vytautas-astrauskas-sensmetry: This suggestion still uses `first`, which was removed. If I try to drop it, test `conflicts_name_root_pins_that_contradict` break:

What goes wrong with "first usage"

A project that lists widget twice in its own file, =1.0.0 and =2.0.0, is the textbook no_solution.
pubgrub reports one incompatibility edge per pin. With the set-matching selection, each edge is mapped
back to the pin that produced it, so the conflict list holds both. With "first usage" both edges map to
=1.0.0, the two conflicts are identical, deduplication collapses them, and the result is:

missing root pin `=2.0.0` in [Constraint { iri: "urn:kpar:widget", constraint: "=1.0.0", required_by:
None }]

A caller then sees a solve that failed over one constraint with nothing to contradict it. The selection
is not a cosmetic tie-break, it is what lets SolveConflict name both sides of a self-contradiction.

andrius-puksta-sensmetry: Yup, my bad, didn't notice.

bindings/py/src/lib.rs
```diff
@@ -488,6 +512,250 @@ fn do_versions_py(
...
+                kind: solver.kind(),
+                conflicts: solver.conflicts(),
+            }
+        }
+        Some(_) | None => Failure::Py(ProjectError::new_err(message)),
+    }
+}
+
+/// `(canonical lockfile text, projects)`, each project converted to a
+/// `LockedProject` dict by `lock::Project`'s `IntoPyObject`.
```
```suggestion
/// Returns `(canonical lockfile text, projects)`
```

andrius-puksta-sensmetry (approved):
Good enough after the nits are fixed, although it would be nice to avoid so much boilerplate for error conversions into python.

core/src/solve/pubgrub.rs
vytautas-astrauskas-sensmetry: In 56bd8c8b7a209195dd4d5af1449282ef766ae32d, I replaced all `enumerate` with a single function that does it internally to make the relationship between the index and dependency set clearer.

core/src/solve/pubgrub.rs
vytautas-astrauskas-sensmetry: Fixed in 1113b2cb0993a31be26265a460c5fcad2cc0da6f. String sorting works for dedup, but potentially messes up the order of reported versions. Therefore, switched to semantic ordering.

core/src/solve/pubgrub.rs
vytautas-astrauskas-sensmetry: Added in 7aea449f23f658732091b3e7048679497bab0324. (I added manually because the comment seems to be now on a wrong position after my latest changes.)

PR #533, opened 2026-09-09
feat(py): `sync()` with an outcome and `no_prune`

vytautas-astrauskas-sensmetry:
This MR adds `sysand.sync(path, *, lock=None, resolution=None, auth=None, provided=None, no_prune=False)`. It installs what the lockfile says into the project's `.sysand` environment, as `sysand sync` does, and returns a `SyncOutcome` listing every project it installed, pruned or kept. When installation fails part-way the `SyncError` carries the outcome so far.

- core: `do_sync` fills a `SyncOutcome` out-parameter (`installed`, `pruned`, `kept`) progressively, so on `Err` the caller holds exactly what happened before the failure.
- sysand (CLI crate): `command_sync` takes the outcome and passes it through. `CliSyncError` names the sync error type for the bindings. Every CLI call site passes a throwaway outcome; CLI behaviour is unchanged.
- py: `sysand.sync` returns a `SyncOutcome` typed dict of `SyncedProject` entries (`iri`, `version`, `path`). A missing lockfile is a `ProjectError`, not an implicit lock. `SyncError.partial` is the outcome so far; environment read or write failures are `EnvError`.
- tests: `do_sync` outcome tests in core over the in-memory environment, `tests/test_sync.py` against the mock index, and five end-to-end scenarios in `test_migration_scenarios.py`.

## Why

A tool driving sysand from Python has to report what a sync changed: which versions went in, which came out, and, when the sync failed half-way, which of those had already happened. `sysand sync` prints progress lines and exits; a caller would have to diff `env.toml` before and after, and after a failure that diff cannot say what the tool did versus what was there before. The CLI also locks implicitly when there is no lockfile, which a tool that has just shown the user a dry-run resolution must never do behind their back.

### `SyncOutcome` (`core/src/commands/sync.rs`)

`do_sync` takes a new argument, `outcome: &mut SyncOutcome`. The outcome holds three `Vec<SyncedProject>` (`iri`, `version`, where `iri` is the entry's first identifier) and a `wrote()` helper that is true when anything was installed or pruned.

The outcome is an out-parameter rather than part of the `Ok` value on purpose: on `Err` the caller still has it, and it contains what had been done up to the failure. Recording happens at the point of each action:

- `kept` gets an entry when the lockfile project is found installed and verified in the environment, or is editable (nothing to install). Provided projects are not recorded; they are not in the environment.
- `installed` gets an entry after a project has been installed.
- `pruned` gets one entry per removed version: the versions not in the lockfile when the IRI is still locked, or every installed version when the IRI has left the lockfile.

**For reviewers:** the prune loop used to iterate `env.uris()` directly, and `LocalDirectoryEnvironment::uris()` yields one entry per installed version of a project (it maps over `env.toml`'s `[[projects]]`). With two versions installed the loop visited the IRI twice and repeated the version scan, its log lines and the `del_uri` call; the second pass found nothing left, so it was harmless. Now that the loop records what it removes, it collects the distinct IRIs first and visits each once. Behaviour is otherwise the same.

### CLI crate (`sysand/src/commands/{sync,add,clone,env,remove}.rs`, `sysand/src/lib.rs`)

`command_sync` grows the same argument and passes it through. `CliSyncError` is the `SyncError` instantiation it fails with, reachable from the returned `anyhow::Error` by `downcast_ref`, in the same way `CliLockError` is for `resolve_lock`.

Every existing caller (`sync`, `add`, `remove`, `clone`, `env install`, `env install --path`) passes `&mut SyncOutcome::default()`. The CLI's behaviour and output are unchanged.

### `sysand.sync` (`_sync.py`, `_model.py`, `lib.rs::do_sync_py`)

```python
outcome = sysand.sync(project_dir, resolution=sysand.Resolution(...))
outcome["installed"]  # [{"iri": ..., "version": ..., "path": "lib/acme_dep_1.0.0"}, ...]
outcome["pruned"]     # [{"iri": ..., "version": ..., "path": None}, ...]
outcome["kept"]       # entries already installed and verified, or editable
```

- `path` is where discovery starts, as for `lock`: the project root and workspace are found by walking up, and the environment is the workspace's or else the project's. The environment directory is created if needed, as the CLI does.
- **No implicit lock.** The lockfile is read before the environment is created. If it is missing the call raises `ProjectError` with `wrote == False` and no `.sysand` directory exists afterwards. This is the one deliberate difference from `sysand sync`: the Python caller is expected to have run `lock` and shown the result first.
- `lock` accepts a `LockResult` from a previous `lock` call (its `text` is used) or lockfile text. This is how a dry run (`lock(..., write=False)`) is installed without ever writing `sysand-lock.toml`.
- `no_prune=True` keeps environment entries that are not in the lockfile, like the CLI flag.
- `resolution`, `auth` and `provided` mean what they mean for `lock`. Provided projects are never installed.
- `SyncedProject.path` is the install directory relative to the environment directory, looked up in the environment metadata after the sync, so `env_dir / path` is where the files are. It is `None` for pruned entries, whose directory is gone.

### Error mapping (`lib.rs`)

Two `Failure` variants are added next to the lock ones, converted to exceptions once the GIL is reacquired:

- `SyncError(message, wrote=..., partial=...)` for a failure inside the sync loop (a download that failed, a checksum mismatch, a missing local source). `partial` is always a `SyncOutcome` dict; `installed` and `pruned` are empty lists when nothing had happened yet. `wrote` is `partial`'s `wrote()`. The `path` entries in `partial` are `None`, because the environment metadata is not rewritten after a failed sync and there is nothing reliable to look them up in.
- `EnvError(message, wrote=...)` for `SyncError::EnvRead` and `EnvWrite`, for a failure to create the environment, and for a failure after the sync loop (merging the lock into the metadata and writing it). `wrote` again comes from the outcome, since installs may have landed before the metadata write failed.

`SyncError` already existed in `_errors.py`; this MR is the first to raise it. Its `partial` docstring said the attribute could be `None` when nothing was written; it is always a dict from a sync failure, and the docstring now says so.

### Tests

Core (`sync_tests.rs`, two tests through `do_sync` over `MemoryStorageEnvironment` with a local-source lockfile entry):

- A first sync reports the entry as `installed` and `wrote()`; a second reports it as `kept` and not `wrote()`.
- An extra installed project is left alone and not reported with `no_prune`, and reported as `pruned` and removed without it.

Python (`tests/test_sync.py`, five tests against the mock index, every call with an explicit `resolution=`):

- Install then keep: the installed entry's `path` points at the files under the environment directory, `env.projects` agrees, and a second sync keeps the same entry.
- Prune unless `no_prune`: a project installed by path is kept with `no_prune=True` and reported as pruned, with `path` `None`, without it.
- No lockfile: `ProjectError`, `wrote == False`, no `.sysand` created. Then `lock(write=False)` followed by `sync(lock=result)` and `sync(lock=result["text"])` install without a lockfile on disk.
- Half-synced: with two dependencies and the second archive answering 500, `SyncError` has `wrote == True`, `partial["installed"]` holds the first dependency only, `partial["pruned"]` is empty, and the message carries the status.
- Authenticated index: without credentials `SyncError` with `wrote == False` and an empty `partial["installed"]`; with `AuthPolicy.bearer` the install succeeds.

### Tests (`tests/test_migration_scenarios.py`)

Five scenarios on top of the `baseline` fixture, with two helpers: `env_versions(baseline)` reads IRI to version out of `env.projects`, and `changed(entries)` turns outcome entries into a set of `(iri, version)` pairs.

- `test_root_pins_library`: the full flow. `versions` shows 0.11.0 and 0.10.3, a dry-run `lock` leaves the lockfile alone, the real `lock` writes the same text, `sync` reports 0.11.0 installed and 0.10.3 pruned, and the environment holds the 0.11.0 files.
- `test_dependent_pins_internally`: a dependent that follows the library with a new release; both new versions are installed and both old ones pruned.
- `test_open_ended_dependent`: a dependent with `>=0.10` is correctly kept at its version while the library moves; `env.projects` is what a caller uses to notice it afterwards.
- `test_authenticated_index`: `versions`, `lock` and `sync` against an index requiring a bearer token, with `AuthPolicy.bearer` and `AuthPolicy.from_env`, and secrets absent from `repr`.
- `test_half_synced`: the archive of the last lockfile entry fails; `SyncError.partial` names the entry that did land, nothing was pruned, and the old 0.10.3 install is still on disk next to it.

The module docstring is updated to mention `sync` and `env.projects`.


core/src/commands/sync.rs
```diff
@@ -378,8 +417,16 @@ where
                 }
             }
         }
+        // `uris()` may yield one entry per installed version; visit each
+        // project once.
```
andrius-puksta-sensmetry: This is unintended. `env.uris()` must return a single entry for every project, not version. Fix it at the source rather than working around it.

bindings/py/src/lib.rs
```diff
@@ -752,6 +789,158 @@ fn do_lock_py(
...
+    dict.set_item("kept", entries(&outcome.kept)?)?;
+    Ok(dict)
+}
+
+/// Typed failure for `command_sync`'s `anyhow::Error`, with what had been
+/// done before it.
+fn sync_error_to_failure(err: anyhow::Error, outcome: SyncOutcome) -> Failure {
+    let message = format!("{err:#}");
+    let wrote = outcome.wrote();
+    match err.downcast_ref::<CliSyncError>() {
```
andrius-puksta-sensmetry: Make the CLI function return the typed error directly.

bindings/py/src/lib.rs
```diff
@@ -752,6 +789,158 @@ fn do_lock_py(
...
+            });
+            (entry.iri.clone(), entry.version.clone(), path)
+        })
+        .collect()
+}
+
+/// A `SyncOutcome` as the `SyncOutcome` typed dict, for `SyncError.partial`.
+/// Install paths are not resolved here: the environment metadata is not
+/// rewritten after a failed sync.
+fn outcome_dict<'py>(py: Python<'py>, outcome: &SyncOutcome) -> PyResult<Bound<'py, PyDict>> {
```
andrius-puksta-sensmetry: Can `IntoPyObject` be derived for `SyncOutcome` instead of this?

bindings/py/python/sysand/_sync.py
```diff
@@ -0,0 +1,75 @@
...
+) -> SyncOutcome:
+    """Install what the lockfile says into the project's ``.sysand``
+    environment, as ``sysand sync`` does, and report every change.
+
+    ``path`` is where discovery starts (see :func:`lock`). Unlike the CLI,
+    a missing lockfile is an error, not an implicit ``lock``: the caller is
+    expected to have seen the resolution first. ``lock`` may be the
+    :class:`LockResult` of a previous :func:`lock` call (its ``text`` is
+    used) or lockfile text; by default ``sysand-lock.toml`` is read from the
+    project root.
```
andrius-puksta-sensmetry:
```suggestion
    used) or lockfile text; by default ``sysand-lock.toml`` is read from the
    project or workspace root.
```

andrius-puksta-sensmetry (changes requested):
Conversions to python types should be more automatic and `downcast_ref` should not be used unless it saves significant boilerplate, as noted above.

core/src/env/local_directory/mod.rs
```diff
@@ -290,12 +290,17 @@ impl ReadEnvironment for LocalDirectoryEnvironment {
...
+        // were installed; list each project's first identifier once.
+        let mut uris: Vec<&str> = Vec::new();
+        for project in &self.metadata.projects {
+            if let Some(uri) = project.identifiers.first()
+                && !uris.contains(&uri.as_str())
+            {
+                uris.push(uri);
+            }
+        }
+        Ok(uris.into_iter().map(|uri| Ok(uri.to_owned())).collect())
```
andrius-puksta-sensmetry, line 303: Don't construct effectively the same vec twice.

core/src/commands/sync.rs
vytautas-astrauskas-sensmetry: Updated.

bindings/py/src/lib.rs
vytautas-astrauskas-sensmetry: Updated

bindings/py/src/lib.rs
vytautas-astrauskas-sensmetry: Updated

core/src/env/local_directory/mod.rs
vytautas-astrauskas-sensmetry: I do not see how to do that without changing the type of `UriIter` to `Vec<String>`.

sysand/tests/cli_env.rs
```diff
@@ -456,6 +457,13 @@ fn env_install_allow_multiple() -> Result<(), Box<dyn std::error::Error>> {
         "project with IRI `urn:kpar:allow-multiple-lib` already has version `1.0.0` installed",
     ));
 
+    // The project is listed once per installed version, not once per
+    // version per `env.toml` entry.
```
andrius-puksta-sensmetry:
```suggestion
    // The project is listed once per installed version
```
Just added confusion. Previous behavior did not make sense, so no need to mention it.

core/src/env/local_directory/mod.rs
vytautas-astrauskas-sensmetry: I could do this change, but it probably deserves a separate PR.

andrius-puksta-sensmetry: Yes, indeed, didn't pay enough attention to types.

andrius-puksta-sensmetry: (no need to change anything then)

andrius-puksta-sensmetry (approved):
Looks good. Found a preexisting bug, which is also nice.

PR #543, opened 2026-09-18
fix(deps)!: let consumers choose the rustls crypto provider

vytautas-astrauskas-sensmetry:
`sysand-core` and `sysand` both took `reqwest` with its `rustls` feature — which is `__rustls-aws-lc-rs` + `__rustls` — and both left reqwest's defaults on, where `default → default-tls → rustls`. Cargo features are additive, so a downstream crate that has already chosen a different provider cannot subtract ours. It gets `aws-lc-rs` as well, and with it `aws-lc-sys`, whose build script needs cmake and, on Windows, NASM, for every target it cross-compiles to.

Four feature edges, from two manifest lines:

```
$ cargo tree -p sysand-py -i aws-lc-rs -e features
reqwest feature "rustls"
  ├── sysand              (sysand/Cargo.toml)
  ├── sysand-core         (core/Cargo.toml)
  └── reqwest feature "default-tls"
      └── reqwest feature "default"
          ├── sysand
          └── sysand-core
```

## What changed

The library names **no** provider. Each crate that ships an artefact names one in its own `default`:

| crate | `default` |
| --- | --- |
| `sysand-core` | `["std"]` — unchanged, no provider |
| `sysand` | `["std", "tls-aws-lc-rs"]` |
| `sysand-py` | `["extension-module", "abi3", "tls-aws-lc-rs"]` |
| `sysand-java` | `["tls-aws-lc-rs"]` — the crate had no `default` before |
| `sysand-js` | untouched; it enables no `networking` |

So what Sysand ships is byte-for-byte the same provider choice as before, while a consumer can write `features = ["networking", "tls-ring"]` and get ring, and only ring.

`sysand-java` is the one that needs `default` for its own sake: `scripts/java-builder.py` builds with no `--features`, so without it every Java build would get `networking` with no provider and fail on the first request.


core/src/resolve/net_utils.rs
```diff
@@ -3,6 +3,13 @@
 
 use std::{error::Error, fmt::Display};
 
+// Both tests assert that a provider is installed, so they only make sense
+// when a `tls-*` feature selected one. Gating on `networking` alone would
+// assert it in `--features std,networking`, which deliberately has none.
+#[cfg(all(test, any(feature = "tls-aws-lc-rs", feature = "tls-ring")))]
+#[path = "./net_utils_tests.rs"]
+mod tests;
```
andrius-puksta-sensmetry: Move this to the end of the file for consistency.

core/src/resolve/net_utils.rs
```diff
@@ -89,7 +96,50 @@ pub(crate) const USER_AGENT: &str = concat!("sysand/", env!("CARGO_PKG_VERSION")
...
+/// spawned thread, so the clone path needs this even though it never touches
+/// [`create_reqwest_client`].
+///
+/// Installing is idempotent by failure. [`rustls::crypto::CryptoProvider::install_default`]
+/// returns `Err` when a provider is already set, and that error is discarded
+/// on purpose, so a provider the application installed first keeps winning and
+/// a second call is harmless.
+///
+/// With neither `tls-aws-lc-rs` nor `tls-ring` enabled this is a no-op, and
+/// installing a provider is the application's job.
```
andrius-puksta-sensmetry, line 109:
```suggestion
/// Install the crypto provider selected by this crate's `tls-*` features as
/// the process-wide rustls default.
///
/// `reqwest` is taken with `rustls-no-provider` so that no consumer is forced
/// onto a particular provider. Every path that can reach a TLS handshake
/// must therefore call this first.
///
/// Only the first call installs the provider, subsequent ones are a no-op.
///
/// With neither `tls-aws-lc-rs` nor `tls-ring` enabled this is a no-op, and
/// installing a provider is the application's job.
```

core/Cargo.toml
```diff
@@ -27,7 +27,19 @@ lenient_checks = []
...
+  "dep:reqwest",
+  "dep:gix",
+  "dep:rustls"
+]  # "dep:reqwest-middleware", "dep:partialzip"
+# Which rustls crypto provider `install_default_crypto_provider` installs.
+#
+# Enabling both is not an error: `--all-features` does it, and
+# `tls-aws-lc-rs` wins, because that is what Sysand ships. With neither, a
+# consumer that enables `networking` must install a provider itself before the
+# first HTTP client is built or the first git fetch runs.
```
andrius-puksta-sensmetry:
```suggestion
# TLS providers. Enabling both is not an error: `--all-features` does it, and
# `tls-aws-lc-rs` wins, because that is what Sysand ships. With neither, a
# consumer that enables `networking` must install a provider itself before the
# first HTTP client is built or the first git fetch runs.
```

core/src/resolve/net_utils.rs
vytautas-astrauskas-sensmetry: Updated.

core/src/resolve/net_utils.rs
```diff
@@ -89,7 +89,48 @@ pub(crate) const USER_AGENT: &str = concat!("sysand/", env!("CARGO_PKG_VERSION")
...
+pub fn install_default_crypto_provider() {
+    #[cfg(feature = "tls-aws-lc-rs")]
+    {
+        _ = rustls::crypto::aws_lc_rs::default_provider().install_default();
+    }
+    // `not(tls-aws-lc-rs)` because the two features are additive and
+    // `--all-features` enables both; aws-lc-rs wins as the shipped default.
```
andrius-puksta-sensmetry, line 116: Add this somewhere (probably in lib.rs) to at least prevent both backends from being included in release builds:
```rust
#[cfg(all(not(debug_assertions), feature = "tls-aws-lc-rs", feature = "tls-ring"))]
compile_error!(
    "`tls-aws-lc-rs` and `tls-ring` must not both be enabled in release builds"
);
```

andrius-puksta-sensmetry: And also if none are enabled:
```rust
#[cfg(not(any(debug_assertions, feature = "tls-aws-lc-rs", feature = "tls-ring")))]
compile_error!(
    "one of `tls-aws-lc-rs` and `tls-ring` must be enabled in release builds"
);
```

vytautas-astrauskas-sensmetry: > Add this somewhere (probably in lib.rs) to at least prevent both backends from being included in release builds:

Including both is legal. It could happen that some project already has both because some other dependencies cannot be unified and Sysand is just another dependency.

If you want, I can add an ad-hoc script that checks that `sysand` binary includes only one of them.

> And also if none are enabled:

That is too limiting. The user may want to use Sysand with their own custom backend that is neither of the two.

core/Cargo.toml
```diff
@@ -27,7 +27,17 @@ lenient_checks = []
...
-networking = ["dep:reqwest", "dep:gix"]  # "dep:reqwest-middleware", "dep:partialzip"
+networking = [
+  "dep:reqwest",
+  "dep:gix",
+  "dep:rustls"
+]  # "dep:reqwest-middleware", "dep:partialzip"
+# TLS providers. Enabling both is not an error: `--all-features` does it, and
+# `tls-aws-lc-rs` wins, because that is what Sysand ships. With neither, a
+# consumer that enables `networking` must install a provider itself before the
+# first HTTP client is built or the first git fetch runs.
```
andrius-puksta-sensmetry, line 38:
```suggestion
# TLS providers. Enabling both is not an error on debug builds: `--all-features`
# does it, and `tls-aws-lc-rs` wins, because that is what Sysand ships. Choosing
# neither is allowed on debug builds, but it is not a working configuration
```

core/Cargo.toml
```diff
@@ -84,7 +94,26 @@ tokio = { version = "1.50.0", default-features = false, features = ["rt", "io-ut
...
 globset = { version = "0.4.18", default-features = false }
-reqwest = { version = "0.13.2", optional = true, features = ["rustls", "stream", "multipart"] }
+# `rustls-no-provider` rather than `rustls`: the latter is
+# `__rustls-aws-lc-rs` + `__rustls`, which would force `aws-lc-rs` on every
+# consumer (see the `tls-*` features). `default-features = false` is needed
+# too, because reqwest's own `default` includes `default-tls`, which *is*
+# `rustls`; `charset`, `http2` and `system-proxy` are the rest of that
+# default, re-named here so dropping it changes nothing else. Losing
+# `system-proxy` in particular would silently stop honouring the OS proxy
+# configuration.
```
andrius-puksta-sensmetry:
```suggestion
# It would be preferable to enable aws-lc-rs/ring on here directly, but no
# feature for `ring` is provided, so we choose `rustls-no-provider` and
# install the chosen provider at runtime ourselves
```

design/architecture.md
```diff
@@ -174,6 +176,15 @@ Each binding crate enables different features on `sysand-core` (all also get
...
 | `sysand-js`    | `js` only (no filesystem or networking — runs in browser) |
 
+`sysand-core` takes `reqwest` with `rustls-no-provider` and names **no**
+crypto provider itself, because Cargo features are additive: one named in a
+library could only be added to downstream, never swapped out. So each crate
+that ships an artefact names a provider in its own `default` —
+`tls-aws-lc-rs` for `sysand`, `sysand-py` and `sysand-java` — and
+`sysand_core::resolve::net_utils::install_default_crypto_provider` installs it
+before the first HTTP client or git fetch. A consumer wanting `ring` enables
+`tls-ring` instead; one wanting neither installs a provider itself.
```
andrius-puksta-sensmetry, line 186:
```suggestion
`sysand-core` takes `reqwest` with `rustls-no-provider` and names **no**
crypto provider itself, because Cargo features are additive: one named in a
library could only be added to downstream, never swapped out. So each crate
that ships an artefact names a provider in its own `default` —
`tls-aws-lc-rs` for `sysand`, `sysand-py` and `sysand-java` — and
`sysand_core::resolve::net_utils::install_default_crypto_provider` installs it
before the first HTTP client or git fetch. A consumer wanting `ring` enables
`tls-ring` instead.
```

core/Cargo.toml
vytautas-astrauskas-sensmetry: > Choosing neither is allowed on debug builds, but it is not a working configuration

As commented above, I think requiring to choose one of the two in release mode is too restrictive for a library.

design/architecture.md
vytautas-astrauskas-sensmetry: > A consumer wanting `ring` enables `tls-ring` instead.

The current implementation allows the user to write their own provider and use it, so this phrasing may misguide them.

sysand/Cargo.toml
```diff
@@ -52,7 +54,7 @@ indexmap = "2.13.0"
 tokio = { version = "1.50.0", default-features = false }
 reqwest-middleware = "0.5.1"
 rpassword = "7.5.4"
-reqwest = { version = "0.13.2", features = ["rustls"] }
+reqwest = { version = "0.13.2", default-features = false }
```
andrius-puksta-sensmetry:
```suggestion
# Features unified with `sysand-core`, see there what's enabled
reqwest = { version = "0.13.2", default-features = false }
```

core/src/resolve/net_utils.rs
andrius-puksta-sensmetry: Ok. Then put these in `main.rs`. Sysand binary is not built when `sysand` is used as a dependency (checked now, it doesn't even catch syntax errors when building `sysand-py`).

andrius-puksta-sensmetry (approved):
With `cfg` compile errors added to `main.rs`, this is good to go. Not a fan of runtime intialization, especially since all paths that could possibly require TLS have to be covered, but I see no better solution.

andrius-puksta-sensmetry:
Hmm, I think `reqwest` can be removed from `sysand` dependencies, since we don't control any of its features there and it's only used to parse URLs, and for that it just re-exports `url` crate, which we already depend on.

sysand/src/main.rs
```diff
@@ -1,6 +1,12 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 // SPDX-FileCopyrightText: © 2025 Sysand contributors <opensource@sensmetry.com>
 
+#[cfg(all(feature = "tls-aws-lc-rs", feature = "tls-ring"))]
+compile_error!("`tls-aws-lc-rs` and `tls-ring` must not both be enabled in release builds");
+
+#[cfg(not(any(feature = "tls-aws-lc-rs", feature = "tls-ring")))]
+compile_error!("one of `tls-aws-lc-rs` and `tls-ring` must be enabled in release builds");
```
andrius-puksta-sensmetry: `not(debug_assertions)` is needed to detect release builds:

```suggestion
#[cfg(all(not(debug_assertions), feature = "tls-aws-lc-rs", feature = "tls-ring"))]
compile_error!("`tls-aws-lc-rs` and `tls-ring` must not both be enabled in release builds");

#[cfg(not(any(debug_assertions, feature = "tls-aws-lc-rs", feature = "tls-ring")))]
compile_error!("one of `tls-aws-lc-rs` and `tls-ring` must be enabled in release builds");
```

vytautas-astrauskas-sensmetry: Any specific reason to check only on release builds?

andrius-puksta-sensmetry: Basically to not break the convenient `cargo check/clippy --all-features`.

PR #545, opened 2026-09-18
feat(py)!: return the CLI's exit code from `_run_cli`

vytautas-astrauskas-sensmetry:
The CLI already produces three distinct exit codes, and the Python binding
collapsed them to two:

```rust
// bindings/py/src/lib.rs
exit_code == ExitCode::SUCCESS
```

`lib_main` returns `err.exit_code() as u8` for a clap failure — **2** for an
unrecognised flag or a missing subcommand — and `ExitCode::FAILURE` (1) for a
runtime error. So:

- `python -m sysand --not-a-flag` exited **1** where the native binary exits
  **2**;
- no Python caller could tell a mistyped flag from a command that ran and
  failed.

## What changed

`lib_main` returns `u8`; `main.rs` wraps it:

```rust
fn main() -> ExitCode {
    ExitCode::from(lib_main(wild::args_os()))
}
```

rather than the other way round. `ExitCode` can be returned from `main` but
not inspected, so an embedder cannot recover the number from it — carrying
the code and wrapping at the edge is the only shape that works. `main.rs` and
the Python binding are its only callers, and the crate is `publish = false`,
so one function beats a near-homonym pair whose `ExitCode` half silently
discards the number.

`run_cli`/`run_cli_blocking` return `u8`; `__main__.py` returns the code.

## Breaking

- `python -m sysand` and the `sysand` console script now exit **2** rather
  than 1 on a usage error, matching the native binary.
- `_run_cli` returns an `int`, so success is falsy for anyone testing it for
  truth. It is underscore-private and absent from `__all__`, but the failure
  mode is silent, hence the `!`.
- `sysand::lib_main` returns `u8` rather than `ExitCode`.


andrius-puksta-sensmetry (approved):
LGTM

PR #546, opened 2026-09-18
feat(cli): render long help as a string

vytautas-astrauskas-sensmetry:
`lib_main` writes help to the process's streams. That is right for a binary
and useless to an embedder that has to *return* the text.

## What changed

```rust
pub fn render_long_help<I, T>(prog: &str, args: I) -> String
```

plus `_render_long_help(prog, args)` in the Python bindings.

**It asks clap rather than reimplementing it.** `--help` is declared global
with `ArgAction::HelpLong`, so appending it reaches whatever subcommand the
arguments name, and clap stops with a `DisplayHelp` error whose `render()` is
the text it would have printed. Walking the `Command` tree by hand instead
would duplicate clap's subcommand matching and, worse, render
`Usage: install` — `bin_name` is assigned during parsing, not construction.

**The program name is a parameter, not `args[0]`.** Callers hold the two
apart: an external command receives the arguments that follow its own name,
with the program name passed separately. It is also what makes usage lines
read `syside project env install`.

**Unparseable arguments return clap's error text**, not a panic and not an
empty string. That is what the existing `ExternalCommand.help`
implementations do — they swallow `SystemExit` and return whatever the parser
wrote — so the consumer's behaviour stays uniform across its subcommands.

**The text is unstyled.** `StyledStr`'s `Display` is colour-unaware by
design. The caller prints it through a writer we know nothing about, so
emitting ANSI escapes here would put them into pipes and files. The binary's
own `--help` is untouched and keeps its colour.


andrius-puksta-sensmetry:
What about logs? They are also sometimes important.

sysand/src/lib.rs
```diff
@@ -134,6 +135,70 @@ where
...
+    }
+
+    let args: Vec<OsString> = args.into_iter().map(Into::into).collect();
+
+    // `--help` is a global argument with `ArgAction::HelpLong`, so appending
+    // it reaches whatever subcommand the arguments name.
+    let argv = iter::once(OsString::from(prog))
+        .chain(args)
+        .chain(iter::once(OsString::from("--help")))
+        .collect();
```
andrius-puksta-sensmetry:
```suggestion
    // `--help` is a global argument with `ArgAction::HelpLong`, so appending
    // it reaches whatever subcommand the arguments name.
    let argv = iter::once(OsString::from(prog))
        .chain(args.into_iter().map(Into::into))
        .chain(iter::once(OsString::from("--help")))
        .collect();
```

bindings/py/tests/render_long_help.rs
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+// SPDX-FileCopyrightText: © 2026 Sysand contributors <opensource@sensmetry.com>
+
+//! `_render_long_help` is what an embedding CLI answers its own `help` with:
+//! clap writes to the process's file descriptors, which a Python-level
+//! stream swap cannot capture.
```
andrius-puksta-sensmetry:
```suggestion
```
Already stated in function docs, no need to repeat.

bindings/py/src/lib.rs
```diff
@@ -142,6 +142,26 @@ fn run_cli_blocking(args: Vec<String>) -> u8 {
     exit_code
 }
 
+/// Clap's long help for the command `args` names, as a string.
+///
+/// For an embedding CLI that must *return* help text rather than let it be
+/// printed: clap writes to the process's file descriptors, which a Python
+/// stream redirect never sees. `prog` is the program name to spell in usage
+/// lines.
```
```suggestion
/// stream redirect of `sys.stdout`/`sys.stderr` does not capture. `prog` is the
/// program name to spell in usage lines.
```

bindings/py/tests/test_cli_help.py
```diff
@@ -0,0 +1,35 @@
+# SPDX-License-Identifier: MIT OR Apache-2.0
+# SPDX-FileCopyrightText: © 2026 Sysand contributors <opensource@sensmetry.com>
+
+"""`_render_long_help` returns clap's long help instead of printing it.
+
+An embedding CLI grafts sysand on as a subcommand and has to *return* its
+help text. clap writes to the process's file descriptors, so redirecting
+``sys.stdout`` — the trick a pure-Python parser allows — captures nothing.
+"""
```
andrius-puksta-sensmetry:
```suggestion
```
No need to repeat.

andrius-puksta-sensmetry (approved):
Looks good, just a few nits.

vytautas-astrauskas-sensmetry:
> What about logs? They are also sometimes important.

Logs are orthogonal to this PR.

PR #547, opened 2026-09-18
fix(py)!: let the binding tree be vendored under another package

vytautas-astrauskas-sensmetry:
Sysand is a library and its Python bindings should support being exposed under a different name than `sysand`. Currently, there are two things that prevent that:

**24 absolute imports across 19 modules** name the top-level package — `import sysand._sysand_core as sysand_rs` and friends.

**The extension resolves its exception classes by module path**, through `pyo3::import_exception!(sysand._errors, …)` ×8. That is wrong in both directions:

- With no `sysand` package present, resolution `panic!`s inside pyo3's lazy type lookup, which lands at the ffi boundary and **aborts the process** on the first error raised. Not a `PanicException` a caller can catch — SIGABRT.
- With a standalone `sysand` *also* installed — an ordinary thing for a user to do — it silently binds **that** installation's classes, so `except syside.project.ProjectError` never matches what the extension raises.

## What changed

Imports are relative. `from . import _sysand_core as sysand_rs` binds the
submodule, so all `sysand_rs.do_*` call sites are untouched, and it resolves
an extension module and a `.py` shim identically.

`_errors.py` hands its classes to the extension at import time via a new
`_register_errors`. The classes stay in Python — the docstrings and the typed
attributes (`wrote`, `conflicts`, `report`, `kind`, `partial`) are why they
are there, and `mypy --strict` runs over that tree.

A `registered_exceptions!` macro keeps every one of the 21 `new_err` call
sites byte-identical. The four that pass keyword arguments move from
`raise_with_kwargs(&X::type_object(py), …)` to `X::new_err_with(py, …)`, and
`raise_with_kwargs` folds into the macro. `use pyo3::PyTypeInfo as _` and the
`same_name_method` allow go with it — the inherent `type_object` they existed
for is gone, and the crate builds with `--deny warnings`.

**Ordering is guaranteed, not hoped for.** Importing any submodule executes
the package `__init__`, which imports `._errors` (line 28) before `._info`
(line 45), the first module that can reach the extension.

**If nothing registered**, the error paths raise `RuntimeError` carrying the
original message plus a note naming `_register_errors`. That state is
reachable — this crate's own Rust tests import the extension bare — so it has
a test rather than an `expect`.


bindings/py/src/lib.rs
```diff
@@ -1150,33 +1138,133 @@ fn do_add_py(path: String, iri: String, version: Option<String>) -> PyResult<boo
...
+    ///
+    /// Reachable only by importing this extension without the Python package
+    /// that owns it, so the message names the step that is missing rather
+    /// than leaving an embedder with a class it has never heard of.
+    fn unregistered(message: &str) -> PyErr {
+        PyRuntimeError::new_err(format!(
+            "{message}\n\nnote: sysand's exception classes have not been \
+             registered with this extension module. Import the Python \
+             package that owns it (its `_errors` module calls \
+             `_register_errors`) before calling into the bindings."
```
```suggestion
            "{message}\n\nnote: sysand's exception classes have not been\n\
             registered with this extension module. Import the Python\n\
             package that owns it before calling into the bindings."
```
Exact implementation details are not important and change over time.

bindings/py/src/lib.rs
```diff
@@ -1150,33 +1138,133 @@ fn do_add_py(path: String, iri: String, version: Option<String>) -> PyResult<boo
...
+
+            /// Hand this extension the exception classes it should raise.
+            ///
+            /// Called by `_errors.py` when the package is imported. The
+            /// classes are *not* looked up by module path, because the
+            /// Python tree is vendored under other parent packages: there
+            /// `sysand._errors` either does not exist, or — when a
+            /// standalone sysand happens to be installed alongside —
+            /// resolves to a different copy of the classes, so
+            /// `except ProjectError` would silently miss.
```
```suggestion
            /// Called by `_errors.py` when the package is imported.
```
Repeated from `_errors.py`.

bindings/py/src/lib.rs
```diff
@@ -1150,33 +1138,133 @@ fn do_add_py(path: String, iri: String, version: Option<String>) -> PyResult<boo
...
+            ///
+            /// The first registration wins; a later one is ignored, since
+            /// each installation carries its own copy of this extension and
+            /// therefore its own state.
```
```suggestion
            ///
            /// The first registration wins. All subsequent ones are ignored, since
            /// the calling module may be initialized multiple times (e.g. `importlib.reload`)
```

bindings/py/src/lib.rs
andrius-puksta-sensmetry: If this is done, `unregistered_error_classes_raise_a_readable_runtime_error` will also need to be updated`.

andrius-puksta-sensmetry (approved):
Just a few nits.

bindings/py/src/lib.rs
andrius-puksta-sensmetry: How is this resolved?

vytautas-astrauskas-sensmetry: I accepted your suggestion and GitHub automatically resolved the thread. Is this what you are asking about?

andrius-puksta-sensmetry: Did not show this to me, even after a forced reload for some reason.

bindings/py/src/lib.rs
vytautas-astrauskas-sensmetry: Updated

bindings/py/src/lib.rs
vytautas-astrauskas-sensmetry: Fixed. It seems that I messed up while signing with PGP.

bindings/py/src/lib.rs
vytautas-astrauskas-sensmetry: Updated.

PR #548, opened 2026-09-18
feat(cli): let an embedder say it does not own the process

vytautas-astrauskas-sensmetry:
`log` allows one logger per process. The CLI installs `env_logger` and the Python bindings install the `pyo3-log` bridge, so in a process that uses both, the second one finds the slot taken. `run_cli` reports that on standard error and continues with the level it was asked for but without its own formatting.

That is right for the binary, where nothing else runs and a taken slot means something unexpected happened. It is noise for an embedder, where it is the arrangement the embedder chose -- and for one whose `sysand` subcommand is a Python entry point it is unavoidable, because importing the bindings is how the CLI is reached at all, so the warning would print on every single invocation.

Add `ProcessOwnership` and the `lib_main_with`/`run_cli_with` pair that takes it. `lib_main` and `run_cli` keep their signatures and stay `Owned`, so the binary is unchanged; `_run_cli` passes `Embedded`, and the note moves to `debug!`, where `-v` still finds it.

The level is set before the branch rather than after, so that note can pass its own filter.


sysand/src/lib.rs
```diff
@@ -209,19 +242,43 @@ pub fn standard_auth_policy(use_credential_store: bool) -> Result<CliAuthPolicy>
...
 }
 
+/// For the process's own binary. An embedder wants [`run_cli_with`].
 pub fn run_cli(args: cli::Args) -> Result<()> {
+    run_cli_with(args, ProcessOwnership::Owned)
+}
+
+/// [`run_cli`], told whether the CLI owns the process. See
+/// [`ProcessOwnership`].
+pub fn run_cli_with(args: cli::Args, ownership: ProcessOwnership) -> Result<()> {
```
andrius-puksta-sensmetry:
```suggestion
fn run_cli_with(args: cli::Args, ownership: ProcessOwnership) -> Result<()> {
```
Nothing calls `run_cli`, and no need to export `run_cli_with`.

bindings/py/tests/test_cli_logger.py
```diff
@@ -0,0 +1,74 @@
...
+# SPDX-FileCopyrightText: © 2026 Sysand contributors <opensource@sensmetry.com>
+
+"""`_run_cli` does not complain about a logger the host installed first.
+
+`log` allows one logger per process. The binding functions install the
+`pyo3-log` bridge and the CLI installs `env_logger`, so in a process that uses
+both the second one loses -- and for an embedder whose `sysand` subcommand is
+reached through Python, the bridge is always first, because importing this
+module is how the CLI is reached at all.
+
```
andrius-puksta-sensmetry:
```suggestion
`log` allows one logger per process. The binding functions install the
`pyo3-log` bridge and the CLI installs `env_logger`, so in a process that uses
both the second one loses.
```

andrius-puksta-sensmetry: Importing the `sysand` and calling `_run_cli` will not install the Python logger.

andrius-puksta-sensmetry (approved):
Looks good, just a couple of nits.

